PR #178, opened 2014-12-02
Add censor rule for FOI Register links

### LINKED ISSUE #177: Prevent private links into the FOI Register site from being published
Links sent to individual users to prompt them to update the status of their request are intended only to be available to that user. A new censor rule to prevent such links from appearing when the email is routed to the for-publication address should prevent someone other than the intended recipient from sending updates.

The current assumption is that the underlying cause of this problem is somewhere in the FOI Register code, although these links could also be published by email text being included in a public reply at some point so a censor rule could be useful in any case.

lizconlan:
Prevents links intended for the originating requestor on the FOI Register site from appearing on WDTK

Closes #177


lizconlan:
Have moved this to the icebox as foi-reg doesn't currently have a live implementation

PR #189, opened 2015-04-14
Corresponding changes for new box style.

crowbot:


garethrees:
New box style being https://github.com/mysociety/alaveteli/pull/2314?

PR #196, opened 2015-05-08
Redesigned footer

wrightmartin:
This adds a new style footer (without type styles) to all pages.

Part of https://github.com/mysociety/alaveteli/issues/1525


garethrees:
Few minor issues to fix but looks great – massive improvement!

wrightmartin:
> Being super picky but the social icons don't look like they're cropped with equal spacing around the edge. Likely to be annoying in the future.

Do you mean the way I've situated the logos inside the coloured squares, or where they are in relation the the links? 

If it's the former, that's by design, due to the portrait orientation of the Facebook logo, if it was equally space around the edges the Facebook logo is tiny.

PR #198, opened 2015-05-19
Code and style homepage hero section - basic styles done

wrightmartin:
Design-wise, this is done, but I'll need a little help with the following bits

- [x] Link up authorities and requests in _frontpage_hero.html.erb
- [ ] <s>Link up 'they have to respond' in _frontpage_hero.html.erb</s>
- [x] Solution for workaround used on https://github.com/mysociety/whatdotheyknow-theme/compare/homepage-redesign-hero?expand=1#diff-097e7e7bd4558fac03bae637f6ec448bR2

Also: Are you happy with the way I've done this - it seemed simplest to override the whole front page in this case


wrightmartin:
screen shot 2015-05-19 at 16 16 25
screen shot 2015-05-19 at 16 16 08

wrightmartin:
Now ready to review again - I didn't need to link up 'they have to respond' in _frontpage_hero.html.erb - my mistake

wrightmartin:
...forgot to push it :)

garethrees:
Not sure what I think, but an option:

screen shot 2015-06-05 at 12 41 58

screen shot 2015-06-05 at 12 42 55

lib/views/general/_frontpage_hero.html.erb
```diff
@@ -0,0 +1,26 @@
...
+        <h2 class="new-request__title"><%= _("Want to know something?") %></h2>
+        <p class="new-request__description"><%= _("Start your own request") %></p>
+        <a class="link_button_green" href="<%= select_authority_path %>"><%= _("Make a request &raquo;") %></a>
+      </div>
+    </div>
+  </div>
+  <div class="hero__learn-more">
+    <div class="row">
+      <div class="learn-more__content">
+        <a href="#learn-more"><%= _("Learn more") %></a>
```
garethrees: I don't think this does anything useful yet. I think we should keep the styling but wait until we have more content on the homepage before we try to link somewhere.

garethrees:
Looks great other than the minor comments

wrightmartin:
> Not sure what I think, but an option:

The only downside is the label on the button won't be the same as the one in the navigation bar now

garethrees:
I think I slightly prefer the wording of the modified version, but I don't mind either way. Happy to go with what's there and change it later if we think it would be better.

PR #213, opened 2015-06-10
[#204] Add API link to the footer

### LINKED ISSUE #204: Add 'API' link to the footer
Related to https://github.com/mysociety/alaveteli/issues/1525

We should encourage people to use it

cc/ @wrightmartin

garethrees:
Fixes #204 

screen shot 2015-06-10 at 12 46 14


garethrees:
Fair increase here, though not massively significant number of actual users:

screen shot 2015-10-20 at 12 22 46

Issue #206, opened 2015-06-08
Locale switcher overlaps logged-in user name when user name is long. 


crowbot:

crowbot:
![screen shot 2015-06-08 at 10 50 50](https://cloud.githubusercontent.com/assets/2912/8032268/b5553fb8-0dcc-11e5-8d5d-648755dafa8b.png)

wrightmartin:
I can fix the reasonable length names, but I'll never come up with a nice solution for names of all length. Can we choose a reasonable point and truncate them?

crowbot:
Were you thinking in CSS or in the template?

wrightmartin:
@crowbot I've done it in CSS for now, feels okay and browser support is v good

crowbot:
Sounds good, thanks. I wasn't sure what browser support was like for truncation these days.

garethrees:
Fixed by https://github.com/mysociety/whatdotheyknow-theme/pull/214

PR #214, opened 2015-06-10
Improve logged in experience

### LINKED ISSUE #205: Improve logged-in nav
https://github.com/mysociety/alaveteli/issues/1525

Made no worse by https://github.com/mysociety/whatdotheyknow-theme/pull/197, but doesn't look as good as the new stuff.

screen shot 2015-06-05 at 11 30 20

wrightmartin:
Closes #205, #206

### DIFF
--- a/assets/stylesheets/responsive/custom.scss
+++ b/assets/stylesheets/responsive/custom.scss
@@ -366,6 +366,22 @@ a.link_button_green_large {
         }
     }
   }
+}
+
+.greeting {
+  //Truncate really long names
+  max-width: 19em;
+  @include respond-min( $main_menu-mobile_menu_cutoff ){
+    max-width: 16em;
+  }
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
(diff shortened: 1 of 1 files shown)


garethrees:
I wonder if we should section up the mobile nav like this:

screen shot 2015-06-11 at 10 15 47

At the moment its not that clear that each group is related to a specific set of actions.

wrightmartin:
@garethrees :+1: yep, I agree

garethrees:
> If it doesn't add anything to the mobile experience, does it add anything to the desktop?

I guess the main point is to show you're logged in as the person you're expecting to be logged in as.

Question is, can we do this any better than what's there now? I expect this is more a question for alaveteli core than the WDTK theme?

wrightmartin:
@garethrees, yes, we should make a ticket there, shall I hide it here for now?

garethrees:
I think we just keep it for now. Lets divide up the nav sections on mobile because that makes sense to do, and then deploy the rest as-is.

PR #215, opened 2015-06-10
Change typeface

wrightmartin:
This switches out the old typeface (Ubuntu?) for mySociety's brand typeface (Source Sans)

### Before
screen shot 2015-06-10 at 16 46 32

### After
screen shot 2015-06-10 at 16 45 37

### DIFF
--- a/assets/stylesheets/responsive/custom.scss
+++ b/assets/stylesheets/responsive/custom.scss
@@ -59,7 +59,7 @@ $action-border: #bf923a;
 /* Selected menu item */
 
 .selected-menu-item {
-  font-weight: bold;
+  font-weight: 600;
   color: $main_menu-active_link_text;
   background-color: $main_menu-active_link_bg;
   &:hover{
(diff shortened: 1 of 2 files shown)


garethrees:
At around 750px wide:

screen shot 2015-06-11 at 11 05 44

garethrees:
Just minor fix to the how it works section, but otherwise :+1: 

I actually prefer Helvetica Neue (or even Arial), but +1 for consistency across sites.

garethrees:
I think this has knocked the button text out of alignment too:

screen shot 2015-06-11 at 14 34 14

wrightmartin:
> I think this has knocked the button text out of alignment too:

I'm not too worried about that, as the next PR will be all-new buttons

wrightmartin:
> 939px:

I can't reproduce it  :(

garethrees:
okay we'll launch it as is and see if we can fix it later

garethrees:
Merged in https://github.com/mysociety/whatdotheyknow-theme/commit/0b162ef70727a7234adfd2ac27b221e08f371950

PR #217, opened 2015-06-11
Redesign buttons

wrightmartin:
This changes the display code for buttons, they are now more consistent with our house style

screen shot 2015-06-11 at 17 03 40
screen shot 2015-06-11 at 17 03 20


garethrees:
Looks great other than needing a bit more contrast for the secondary buttons. It also includes a few commits too many. I'll sort that out when I merge it.

Issue #212, opened 2015-06-10
Redesign homepage


garethrees:
Taken over from https://github.com/mysociety/alaveteli/issues/1525

## Metadata

- **Team:** @wrightmartin, @garethrees
- **Prototypes:** http://invis.io/D52MTYYTK
- **Final Mockup:**

![wdtk-home](https://cloud.githubusercontent.com/assets/2292925/7477385/58a95ccc-f34b-11e4-92e8-a2e49f2dcf8f.png)

## Todo

- [x] Improve mySociety Banner (PRs: #194)
- [x] Improve Banner (PRs: #197)
- [x] Improve locale switcher (PRs: #195)
- [x] Improve logged-in nav (Issues: #205, #206), (PRs: #214)
- [x] Improve search box (PRs: https://github.com/mysociety/alaveteli/pull/2273)
- [x] Add hero section (PRs: #198, #210)
- [x] Add how it works section (PRs: #211)
- [x] Improve latest requests listing (PRs: #199)
- [ ] Add user testimonials (Issues: https://github.com/mysociety/alaveteli/issues/2383)
- [x] Improve footer (PRs: #196, #201)
- [x] Improve site-wide fonts (PRs: #215)
- [x] Change primary colour (Red -> Blue) (PRs: #216)
- [x] Tidy up sidebars
- [x] Final polish (Issues: #218, #230, #222) 
- [x] Add note to analytics when launched
- [ ] Refactor `custom.scss` (Issues: #208 ~~https://github.com/mysociety/alavetelitheme/issues/48~~)
- [x] Update `print.scss` (Issues: #223), (PRs: #224)

---

- [x] Blog post / announcement? (Issues: #229)
- [ ] Change other promo material to blue (Issues: #207, #235)
- [ ] Update screenshots on mysociety.org, alaveteli.org, etc (Issues: https://github.com/mysociety/alaveteli/issues/2777)
- [ ] Generate translations (Issues: #234)

MyfanwyNixon:
I've drafted a blog post ready for launch, and would be glad if someone could fact-check it. 
https://blogs.mysociety.org/mysociety/?p=24657&preview=true

* Parts where I particularly need confirmation (because I made assumptions):

> like almost everything we do these days, the changes will be subjected to scrutiny from our Research team.

> As with all site improvements, the design will be rolled out as an option on the wider Alaveteli codebase, so that it can be implemented by anyone running an Alaveteli site.

* Also:

[...]

PR #220, opened 2015-06-12
Sidebars now look like the reference design

wrightmartin:
Improved spacing, typography and button layout

### Before
screen shot 2015-06-12 at 09 44 36

### After
screen shot 2015-06-12 at 09 44 42

### DIFF
--- a/assets/stylesheets/responsive/custom.scss
+++ b/assets/stylesheets/responsive/custom.scss
@@ -1290,3 +1290,8 @@ dd {
     color: #777;
   }
 }
+
+.act_link {
+  line-height: 1.5em;
+  margin-bottom: 0.5em;
+}
--- a/assets/stylesheets/responsive/custom.scss
+++ b/assets/stylesheets/responsive/custom.scss
@@ -1268,3 +1268,17 @@ dd {
     }
   }
 }
+
+.request_short_listing {
+  h3 {
+    font-size: 1.125em;
+  }
+  p {
+    font-size: 0.875em;
+    line-height: 1.5em;
+  }
+  time {
+    display: block;
+    color: #777;
+  }
+}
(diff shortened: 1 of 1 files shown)


garethrees:
I think the small headings have made it less clear. They don't divide the sections as well, so it all looks like one long sidebar. I think the headings need to be a bit bigger and bolder (I like the style, though!) and the rest of the text needs to be smaller (in proportion to the headings).

garethrees:
> Sidebars now look like the reference design

Is there a specific design mockup for the sidebars anywhere, or do you just mean in terms of fitting in with the homepage mockup? (#212)

wrightmartin:
@garethrees here's the design this comes from

screen shot 2015-06-12 at 12 09 04

wrightmartin:
Improvement?

screen shot 2015-06-12 at 14 01 21

garethrees:
Definitely better, but I think the problem still exists.

Maybe like this?

screen shot 2015-06-12 at 14 34 08

I still think it looks a bit busy, but I think the font proportions are better.

wrightmartin:
The trick with that type of heading design (capped up) is too keep it smaller than the body text, it looks neat. As soon as it gets bigger than or equal to, it should be normal case to stop it looking clumsy.

garethrees:
> The trick with that type of heading design (capped up) is too keep it smaller than the body text, it looks neat. As soon as it gets bigger than or equal to, it should be normal case to stop it looking clumsy.

Yeah, I don't disagree. I just think that in the current sidebar the small headings get totally lost.

garethrees:
Okay so we could start from the point of tidying up the existing related request listing:

- make them a bit smaller
- small grey date to match homepage
- display block on the dates to always put them on a newline

screen shot 2015-06-12 at 14 59 10

wrightmartin:
> Okay so we could start from the point of tidying up the existing related request listing:
>
> - make them a bit smaller
> - small grey date to match homepage
> - display block on the dates to always put them on a newline

Yep, a fine idea

screen shot 2015-06-12 at 19 15 09

PR #216, opened 2015-06-11
Change colour scheme

wrightmartin:
This updates the colour scheme of WDTK to the new one, using mySociety colours

### Before
screen shot 2015-06-11 at 11 19 52

### After
screen shot 2015-06-11 at 11 19 46

PR #227, opened 2015-06-19
Style pagination

wrightmartin:
### Before

screen shot 2015-06-19 at 14 23 30

### After

screen shot 2015-06-19 at 14 24 03

### DIFF
--- a/assets/stylesheets/responsive/custom.scss
+++ b/assets/stylesheets/responsive/custom.scss
@@ -1447,3 +1447,37 @@ a img.alignleft {
     margin: 0;
     padding: 0 0.33em 0.33em;
 }
+
+// Pagination
+div.pagination {
+  // using div.pagination here to override specificity
+  padding-top: 1em;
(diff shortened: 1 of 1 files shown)


assets/stylesheets/responsive/custom.scss
```diff
@@ -1229,3 +1253,37 @@ dt {
...
+}
+
+div.pagination span.current,
+div.pagination a,
+div.pagination a:visited,
+.pagination .previous_page,
+.pagination .next_page,
+.pagination .current {
+  @extend .button-secondary;
+  margin: 0 0.2em;
```
garethrees: Need a hover/focus state on these

wrightmartin: There is - it uses the .button-secondary one, probably hard to spot, due to the bugs we talked about earlier that are solved in #216

garethrees: :+1:

Issue #218, opened 2015-06-12
Cosmetic tweaks to follow new colour scheme


wrightmartin:
Final polish things that can be in one PR (can't be done until everything from #212 is merged into master)

- [x] Double check colour variables are being used in custom.scss
- [x] Slightly darken the nav bar
- [x] Filtering interface colours need softening
- [x] Drop shadows on correspondence
- [x] Vary button styles on Authority page
- [x] Pagination styles (bottom of Authority page)
- [x] Button at bottom of authority list on homepage
- [x] Pagination on mobiles is messy 
- [x] navimg/logo-trans-small.png needs to be updated so that HTML versions of attachments show the new logo.

garethrees:
Pagination needs a tweak on mobile and on some desktop pages:

Mobile:

![screen shot 2015-06-22 at 14 57 41](https://cloud.githubusercontent.com/assets/282788/8286567/3ac021d2-1902-11e5-89ab-24c34c176269.png)

Desktop:

https://www.whatdotheyknow.com/list/successful?commit=Search&page=9&query=software&request_date_after=&request_date_before=&utf8=%E2%9C%93

![screen shot 2015-06-22 at 17 14 21](https://cloud.githubusercontent.com/assets/282788/8286566/3695df2a-1902-11e5-8ffc-4f2ee6bdf715.png)

PR #233, opened 2015-06-22
visual tweak to correspondence thread, part of #218

wrightmartin:
Before

screen shot 2015-06-22 at 15 47 41

After

screen shot 2015-06-22 at 15 47 49


garethrees:
I don't think this brings WDTK nearer to any of our other sites.

wrightmartin:
So two things here

- The change to the shadow adds more contrast and makes the correspondence look sharper. It catches the eye better and helps bring it out of the rest of the page, as the thing you should be looking at
- The rounded corners brings it inline with the rest of the site, take a look around and you'll notice there's no sharp corners anywhere.

PR #243, opened 2015-07-02
pagination now behaves nicer in longer lists and smaller screens

wrightmartin:
screen shot 2015-07-02 at 16 55 48

### DIFF
--- a/assets/stylesheets/responsive/custom.scss
+++ b/assets/stylesheets/responsive/custom.scss
@@ -1541,7 +1541,7 @@ div.pagination a:visited,
 .pagination .next_page,
 .pagination .current {
   @extend .button-secondary;
-  margin: 0 0.2em;
+  margin: 0 0.2em 0.5em;
 }
 
 .pagination .current {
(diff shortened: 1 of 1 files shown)


garethrees:
The overlap is not ideal.

You can [customise](https://github.com/mislav/will_paginate/wiki/API-documentation#will_paginatecollection-options) the number of pages shown around the current page (see `:inner_window` and `:outer_window`), but its more faff than its worth to override.

Would it be worth doing this if there are high number of pages?

```ruby
page_opts = if collection.count / @per_page >= 10
  # Reduce the windows when there are more than 10 pages
  { :inner_window => 1, :outer_window => 2 }
else
  # Use the default
  { :inner_window => 2, :outer_window => 4 }
end
```

In any case, :+1:  on this.

wrightmartin:
> Would it be worth doing this if there are high number of pages?

Yep, if it works we should look into it - something for Alaveteli core though?

garethrees:
> Yep, if it works we should look into it - something for Alaveteli core though?

Yeah too much of a faff to override here.

PR #244, opened 2015-07-02
replace logo with new colourway

wrightmartin:


garethrees:
:hotsprings: :+1:

garethrees:
Bah just merged this but noticed the file size has doubled!

```
assets/images/navimg/logo-trans-small.png | Bin 4505 -> 9644 bytes
```

@wrightmartin anything you can do about that? http://kraken.io?

wrightmartin:
Yeah I'll recompress it and push again

garethrees:
That would be an awesome check to have. We don't actually run CI on themes (https://github.com/mysociety/alaveteli/issues/1853) but might still be worth adding to alaveteli itself

PR #260, opened 2015-09-15
Issues/246 update favicons

### LINKED ISSUE #246: Blue favicon
To match new theme

wrightmartin:
This adds a new favicon, and also icons for Androin, IE11 and iOS.

NOTE: favicon.ico is committed here but isn't in the correct place, does it need manually placing on the server?

Closes #246


garethrees:
> NOTE: favicon.ico is committed here but isn't in the correct place, does it need manually placing on the server?

It just uses [the default](https://github.com/mysociety/alaveteli/blob/develop/app/assets/images/favicon.ico) at the moment, so yeah we need to commit `favicon.ico` here, as you've done :+1:

PR #287, opened 2015-12-17
[#625] Extract school logic

garethrees:
Fixes mysociety/alaveteli#625
Requires https://github.com/mysociety/alaveteli/pull/2963


garethrees:
Merge in https://github.com/mysociety/alaveteli/issues/2941

PR #305, opened 2016-03-18
[#298] Request Made Conversion Tracking

### LINKED ISSUE #298: Add Adwords conversion tracking code
See https://github.com/mysociety/Non-Site-Specific/issues/687

As a condition of the higher level Google Adwords grant, we need to be tracking at least one goal. The sooner we can get this in place, the sooner I can reapply and the sooner we'll get more users back!

@jacksonj04 says:

> https://support.google.com/adwords/answer/6095821 is the guide on doing it.

> Basically, on each site, on each page which counts as a 'conversion' (eg the WTT success page, the WDTK success page, the TWFY postcode search result page, the FMS success page) we add a bit of code which tells AdWords "whatever you did, worked". When AdWords spots an incoming "success" message, it records that against the relevant ad (assuming the user arrived on the site via an ad) as a conversion.

For WDTK, yes, the obvious goal page would be the 'you've sent your request' confirmation page. Could also consider signing up for an alert to be a goal.

garethrees:
Connects to #298

### DIFF
--- a/lib/views/request/_request_sent.html.erb
+++ b/lib/views/request/_request_sent.html.erb
@@ -68,3 +68,22 @@
     </div>
   </div>
 </div>
+
+<!-- Google Code for WDTK Request Made Conversion Page -->
+<script type="text/javascript">
+  /* <![CDATA[ */
+  var google_conversion_id = 1067468161;
+  var google_conversion_language = "en";
+  var google_conversion_format = "3";
+  var google_conversion_color = "ffffff";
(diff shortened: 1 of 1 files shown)


garethrees:
FWIW, I did consider making this a partial in alaveteli core, but didn't for two reasons:

- We can get this deployed quicker
- We might want to start tracking other conversions (comments, signups, follows, etc) so it would be better to release them all as an individual feature set that we can ship with good documentation on how to set it all up.

lizconlan:
👍 looks good to me!

Issue #309, opened 2016-04-20
Insert donation CTAs at sweet spots in WDTK user flow


MyfanwyNixon:
Moving across to individual repos from mysociety/sysadmin#679

Primarily a design issue: latest suggestion from @zarino  is a test where we display donation links only at each site’s pleasure points:

-     Work out where the pleasure points are in the user flows for our products (I think @MyfanwyNixon did some research on this before?) and put donation links there, with Analytics tracking IDs.
-     Make the donation links really persuasive. Think Save The Children or RSPCA level heart-tugging. Get crying children on there. Again, all-out, no shame.
-     Put them up for exactly 2 months, as before.
-     See whether donations rise.

wrightmartin:
I think this is an obvious one, but here's a first shot at adding donate to the confirmation screen

![screen shot 2016-05-17 at 15 59 40](https://cloud.githubusercontent.com/assets/2292925/15327205/856323d4-1c48-11e6-81b9-6b919f04146f.png)

garethrees:
I'd argue it needs to be even bolder – I'd probably scan over that.

zarino:
I've said it before, I'll say it again: **go all out.** I believe my words last time were "crying children", I wasn't joking. If this thing is worth doing, it's worth going the whole hog. We should stop short of misleading people, but we shouldn't be afraid of tugging at their emotions.

wrightmartin:
Okay, maybe we need a two pronged approach though,  @MyfanwyNixon fancy taking a crack at the copy?

MyfanwyNixon:
A picture of a crying MarkC it is, then.

Or if you prefer the power of words, hmm, there are many avenues we could go down. We often end up talking about the '£5 helps us run for x days' or 'powers a server for a month' or whatever. Those figures can be hard to pin down but they could be persuasive.

It's also tempting to go with a message like 'we need your donations to keep the site running' but that's not strictly true, is it.

## Help keep WhatDoTheyKnow running ##
Your donations help us make Freedom of Information available to all. DONATE NOW

## Help us make FOI accessible to all ##
WhatDoTheyKnow is a project of the charity mySociety, run by volunteers. DONATE NOW

zarino:
More than making FOI accessible, I'd argue lots of the stuff mySociety has done has helped *protect* FOI provision from being weakened, especially in the UK. If this were a 38 Degrees web page, they'd be saying something like:

## Your right to FOI is under attack around the world

[...]

PR #318, opened 2016-06-02
Add donate prompts

wrightmartin:
Update
----------

To do 
- [x] Whitespace fix
- [x] https://github.com/mysociety/internal/issues/11

Part of #309

Here's how the messages look now

screen shot 2016-06-01 at 17 24 05
screen shot 2016-06-01 at 17 30 15

A note for the future, the way the notices and messages are put together in Alaveteli is messy and quite hard to follow - I'll open a ticket to refactor these

### DIFF
--- a/assets/stylesheets/responsive/custom.scss
+++ b/assets/stylesheets/responsive/custom.scss
@@ -322,12 +322,11 @@ a.link_button_green_large {
 /* Notices, errors */
 
 #notice, #error, .errorExplanation, #request_header_text, #hidden_request, .undescribed_requests, .describe_state_form form {
-  border-top: 10px solid;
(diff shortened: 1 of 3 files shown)


wrightmartin:
(forgot to note - this was using the 'static-states' branch of alaveteli core https://github.com/mysociety/alaveteli/tree/static-states)

MyfanwyNixon:
Ah yes, I see your point @crowbot.
How about something like:

> Your donations help us make Freedom of Information more accessible, for everyone.

Bit tricky getting so many concepts into a short message! And it'd be quite tempting to take a Google Ads-style approach and rotate several variants until you see which one brings the best results..

PR #322, opened 2016-06-16
Remove exception template override.

crowbot:
Once the changes made in 416ad02a6d01b2e6cdbcc94e7bd143c9c21e59c6
are made here, it would be identical to the template in core.


lizconlan:
:+1: yay, fewer overrides!

lizconlan:
looks as though mysociety/alaveteli@416ad02a6d01b2e6cdbcc94e7bd143c9c21e59c6 has made it into develop, not sure whether it's been deployed to wdtk

garethrees:
The view still exists in master: https://github.com/mysociety/whatdotheyknow-theme/blob/master/lib/views/general/exception_caught.html.erb. Looks like this got closed without merging?

garethrees:
Ah cool 👍

PR #323, opened 2016-06-20
Contact page

zarino:
Connects to mysociety/alaveteli#74.

**To-do:**

- [x] Reject form unless "I understand" checkbox has been checked.
- [x] Form error handling – validation errors should be displayed in the correct place, the last `houdini-input` should be "checked" so the form is visible, and the user should be sent down to the `#wdtk` anchor.
- [x] Hide "Contact" section of sidebar.
- [ ] Welsh translation.
- [x] Nicer error message for not checking the box

screen shot 2016-06-20 at 12 31 02

screen shot 2016-06-20 at 12 31 07

screen shot 2016-06-20 at 12 31 16

### DIFF
--- a/lib/views/help/_contact_form.html.erb
+++ b/lib/views/help/_contact_form.html.erb
@@ -57,6 +57,12 @@
     <% end %>
 
     <div class="form_button">
+        <script><!--
+            if (!!navigator.userAgent.match(/Version\/[\d\.]+.*Safari/)) {
+                $('#contact_email').removeAttr('required');
+                $('#contact_understand').removeAttr('required');
+            }
+        //--></script>
         <%= hidden_field_tag(:submitted_contact_form, 1) %>
         <%= submit_tag "Send message to WhatDoTheyKnow volunteers", :disable_with => "Sending..." %>
     </div>


lizconlan:
@zarino - could you check I've not broken your design before I hand this over for a review of the Rails code?

lib/views/help/_contact_form.html.erb
```diff
@@ -0,0 +1,64 @@
...
+            government, and the WhatDoTheyKnow volunteers <strong>cannot</strong>
+            help me with personal matters relating to government services.
+        </label>
+    </p>
+
+    <% if not @user %>
+        <p>
+            <label class="form_label" for="contact_name">My name:</label>
+            <%= f.text_field :name, :size => 20 %>
+            (or <%= link_to "sign in", signin_url(:r => request.fullpath) %>)
```
zarino: @lizconlan – This return path should include the "#wdtk" location fragment, so that when they return to the contact page after logging in, they're automatically taken down to the right section of the page.

zarino:
@lizconlan – I've added a few commits tidying up the error styles, but other than that, it all looks good!

lizconlan:
@zarino I got a bit confused (easily done, admittedly) when there were "required" form fields as it just seemed as though the form was broken

lizconlan:
Have fixed the post signin redirect. But... the client-side validation doesn't work in Safari, it just looks like the form submit's broken.

zarino:
Sigh. Safari (on Mac and iOS) doesn't show a notification when a required form input is left empty: http://caniuse.com/#search=required ([webkit bug](https://bugs.webkit.org/show_bug.cgi?id=28649) from 7 years ago).

It is, however, beneficial for users with more compliant browsers, because it saves the page having to reload before you see your mistake.

Maybe we use JavaScript to remove the "required" attribute on Safari only? Would the form submit then?

lizconlan:
7 year old bug in Safari, you say? Wish I could be more surprised. Yeah that should work - the only other thing I can think is to make the disabled submit button more obvious but that's probably harder/more disruptive.

zarino:
Yes, we should aim to avoid the submit button being disabled at all. If that means we have to remove form validation entirely for all browsers, just so that it works for Safari, then so be it. But it'd be nice if we could just remove it for Safari, and keep it for the more modern browsers.

lizconlan:
A bit clunkily done - if I say so myself - but should be Safari-proof now

lib/views/help/contact.html.erb
```diff
@@ -1,110 +1,97 @@
...
+        </label>
+    </h2>
 
+    <input class="houdini-input" type="radio" name="goals" id="goal1">
+    <div class="houdini-target contact-page__options">
+        <ul>
+            <li>
+                If you want to make a public Freedom Of Information
+                request to a UK government department, you can
+                <a href="<%= new_request_path %>">start here</a>.
```
garethrees: Clicking this redirects you to the homepage, You want `select_authority_path`.

lib/views/help/contact.html.erb
```diff
@@ -1,110 +1,97 @@
...
-
-    <h1>Contact the WhatDoTheyKnow team</h1>
-    <% if !flash[:notice] %>
-            <ul>
             <li>
-            Please read the  <a href="<%= help_about_path %>">help page</a> first, as it may
-            answer your question quicker.
+                If you want to make a public Freedom Of Information
+                request to a UK council or public authority, you can
+                <a href="<%= new_request_path %>">start here</a>.
```
garethrees: Clicking this redirects you to the homepage, You want `select_authority_path`.

lib/model_patches.rb
```diff
@@ -135,4 +135,11 @@ def alert_new_response_reminders_with_alert_survey
             alias_method_chain :alert_new_response_reminders, :alert_survey
         end
     end
+
+    ContactValidator.class_eval do
+        attr_accessor :understand
+
+        validates_acceptance_of :understand,
+                                :message => N_("Please check the box to confirm your understanding")
```
garethrees: "Please confirm that you understand that WhatDoTheyKnow is not run by the government, and the WhatDoTheyKnow volunteers cannot help you with personal matters relating to government services."?

garethrees:
Couple of minor bits but after that good to go. Lets handle the Welsh translations as part of https://github.com/mysociety/alaveteli/issues/3137, as there'll be some additional translation to do there.

garethrees:
Assigned to @lizconlan to handle minor fixups and merge

lib/views/help/contact.html.erb
lizconlan: D'oh! Thanks

lizconlan: (ah, that's been copied across from the original and is what's currently on the site - good catch @garethrees)

lib/model_patches.rb
lizconlan: 👍

PR #325, opened 2016-06-27
add UTM tags to mysociety footer

wrightmartin:


jacksonj04:
My view would be along #318, that `utm_source` should equal the site name, `utm_campaign` should reflect the 'why' of the link being there (specific campaign, generic CTA etc) and the `utm_medium` being reflective of where in the site the link was.

I think the medium is the trickiest part to get right, other than things like `footer` which is pretty uncontestable. I'd imagine things like `confirmation_page` or `signup_page` would be in order, but this leads to trying to define a common vocabulary between sites.

garethrees:
Also https://github.com/mysociety/whatdotheyknow-theme/commit/f6a04910c661120c0eec366f7541478756c50b77

wrightmartin:
@zarino bringing you in here because you proposed the tags? Am I right in assuming we're trying to make all the footer tags consistent across all sites?

zarino:
The original plan was as described in the table on https://github.com/mysociety/standard-footer/issues/9.

Sounds like @jacksonj04 largely agrees – although he's suggesting loading the `utm_medium` with more meaning than just `"link"` or whatever, and instead using it to say *where* the link was. I guess that's fair enough, as long as it can be consistent across all the sites.

Does that help?

jacksonj04:
I'm fine with `utm_medium` being link/banner/etc and `utm_source` reflecting the _where_. Like @zarino said, as long as we're consistent then everything is shiny.

garethrees:
I've used `source` as the site name, e.g. `whatdotheyknow` and `utm_content` for the where on the site e.g. `sitebar_similar_requests`. 

- https://github.com/mysociety/alaveteli/pull/3354/commits/6f0eb949cbf5613164147340bd64cbefd8f33207#diff-9adef49c1b51a16e3d16877523eed5b6R62
- https://github.com/mysociety/whatdotheyknow-theme/commit/f6a04910c661120c0eec366f7541478756c50b77

Does that mean we want to change those?

jacksonj04:
Gah, sorry, `utm_content` reflecting the where, not `utm_source` (which should _definitely_ be the site).

PR #330, opened 2016-07-19
Changes for Release 0.25.0.0 compatibility

lizconlan:


lib/views/general/_frontpage_bodies_list.html.erb
```diff
@@ -17,7 +17,7 @@
 
     <p>
       <strong>
-        <%= link_to _('Browse all authorities &rarr;'), list_public_bodies_default_path, :class => 'button-secondary' %>
+        <%= link_to _('Browse all authorities...'), list_public_bodies_default_path, :class => 'button-secondary' %>
```
garethrees: The `&rarr;` is intentional here to match up with the "What information has been released?" section https://www.whatdotheyknow.com/#frontpage_examples

lizconlan: Ah, cool - wasn't sure (that's why I left it as a fixup). Thanks!

PR #335, opened 2016-08-05
Add Act vs Share experiment

garethrees:
Part of https://github.com/mysociety/alaveteli/issues/3394
Requires https://github.com/mysociety/alaveteli/pull/3437

Defaults to showing the original if JS not available


lizconlan:
👍 should do the trick 🍰

PR #337, opened 2016-08-05
[#3394] act writetothem

garethrees:
Part of https://github.com/mysociety/alaveteli/issues/3394

screen shot 2016-08-05 at 16 24 21

Adds event tracking to Write To Them link


lizconlan:
👍 looks good, shame the insert-new-link-here bit has to be so complicated!

garethrees:
> shame the insert-new-link-here bit has to be so complicated

Yeah, but we can fix that if we keep the variant, so not too concerned 🍦

Issue #339, opened 2016-08-10
Remove mention of WhatDoTheyKnow for Councils


BenJam:
We just recived a reqeust about this, probbaly best to remove this reference  https://www.whatdotheyknow.com/help/privacy#real_name

crowbot:
There are still anonymous requests on the site from this integration though - perhaps we should rephrase it into the past tense instead?

PR #353, opened 2016-11-14
[#339] Make foi-register note past tense

garethrees:
Missed in https://github.com/mysociety/whatdotheyknow-theme/pull/342


lizconlan:
👍

PR #357, opened 2016-12-01
Issues/alaveteli pro 60/pro navbar

crowbot:


lib/views/general/_responsive_header.html.erb
```diff
@@ -9,7 +9,7 @@
     </div>
   <% end %>
 <% end %>
-<div id="banner" class="banner <%= @pro_area ? 'pro' : '' %>" role="banner">
+<div id="banner" class="banner <%= @pro_area ? 'banner--pro' : '' %>" role="banner">
   <div id="banner_inner" class="banner_inner">
     <div id="banner_content" class="banner_content">
       <div class="only-show-for-print">
```
wrightmartin: ooh, no, that should be in core too actually. Thanks for spotting that 👍

PR #361, opened 2016-12-12
Reconcile with alaveteli-core develop

garethrees:
Needs some work on the request action menu – see todos in 04231eb.

screen shot 2016-12-12 at 18 40 06


garethrees:
Current progress – not sure why the button margins etc are so different:

screen shot 2016-12-14 at 12 18 27
screen shot 2016-12-14 at 12 18 34

garethrees:
Height miss-match can be fixed with:

```css
line-height: 24px;
```

garethrees:
Here's how it looks in ad7f808

screen shot 2016-12-14 at 13 14 34

Issue #284, opened 2015-12-15
'Donate' banner


MyfanwyNixon:
As suggested by @markcridge:
Can we display a banner across the top of the site, from Monday to Weds next week, and linking to https://www.mysociety.org/donate/

> Help us make websites that empower people. Your donations support sites like this.

crowbot:
@lizconlan - as you'll be the only alaveteli dev around, is this something you feel confident doing?

lizconlan:
@crowbot sounds like something I should be able to cope with, any gotchas I need to be aware of?

crowbot:
Shouldn't be - it's just a case of editing
`lib/views/general/_popup_banner.html.erb' here on the master branch, and
then redeploying the site.

lizconlan:
thanks @crowbot! (note to self, appears to need a css tweak as well as the parent container has a hardcoded `display:none` (inherited from outside the theme?)

crowbot:
Shouldn't need css changes - the `display:none` will be overriden.

lizconlan:
\*clears cookies\* aha, that's the culprit!

lizconlan:
@MyfanwyNixon does this look right? have made the entire text a link

![screen shot 2015-12-17 at 10 22 31](https://cloud.githubusercontent.com/assets/27760/11867337/6d7b8ec2-a4a8-11e5-9a9a-d40657fb79a2.png)

MyfanwyNixon:
I think maybe just the second sentence as a link. Thanks!

lizconlan:
@MyfanwyNixon cheers, will fix :)

lizconlan:
will now look more like this:

![screen shot 2015-12-17 at 10 28 20](https://cloud.githubusercontent.com/assets/27760/11867449/f8dacb5e-a4a8-11e5-9f3a-c8b9a1242e84.png)

MyfanwyNixon:
:+1: :+1:

lizconlan:
Ready to go, will set myself a reminder to merge it into master and deploy on Monday (and take it down Thursday morning, if that's ok?)

MyfanwyNixon:
Either sounds good - Mark originally said 'Monday to Wednesday', however I am sure first things Thurs would be fine i it's convenient. Thanks.

lizconlan:
[...]

PR #362, opened 2016-12-14
Add donate banner

lizconlan:
Christmas/New Year donation banner

Connects to #284


garethrees:
@lizconlan if you can hold off deploying this until I've merged #361 that would be great! :)

garethrees:
too late… never mind :)

PR #369, opened 2017-01-03
Dec 2016 design polish

wrightmartin:
Tiny tweaks to design of new features that went live in 2016

### DIFF
--- a/assets/stylesheets/responsive/custom.scss
+++ b/assets/stylesheets/responsive/custom.scss
@@ -826,7 +826,27 @@ input#reedit_button {
 }
 
 /* Authority page */
+.authority__header {
+  border: 0;
+}
+
+.authority__header__action-bar-container {
+  background-color: #fff;
+  width: 100vw;
+  margin-left: calc(-1 * ((100vw - 100%) / 2));
+  margin-bottom: 1em;
+  margin-top: 1em;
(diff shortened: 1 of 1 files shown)


garethrees:
I've extracted 34fcc84 and aeaec1b out of this PR and got them released as 77f5a1b0f0682cc8eed04c58c357ce67ad0a0fcd and ab2a7c893b482685ab5ea3f29fabe13a5fcbc8a2.

I didn't merge the remaining commits because I'm not sure about the white background of the action menus. I feel it breaks the flow of the page (which may be intentional?) and the colour makes it look a bit out of place.

screen shot 2017-01-13 at 16 47 16

If I was going to do this I'd have thought a subtly darker shade of the background colour blends in better, but I'm still not convinced at the full width:

screen shot 2017-01-13 at 16 52 25

IDK, it feels like the request title is jammed in between them and looks a bit cluttered.

wrightmartin:
You're right @garethrees that does look better darker. Have a look at a request page with a long title it doesn't look quite as jammed in

garethrees:
We decided that we won't introduce new changes around the action menu section given how rapidly pro is changing. We'll revisit this after pro settles down and we see if there are things we can merge in to the general user experience. Cheers!

PR #389, opened 2017-03-31
Issues/alaveteli pro 201/reminders of good requesting

crowbot:


lib/views/alaveteli_pro/info_requests/_new_request_advice.html.erb
```diff
@@ -0,0 +1,56 @@
+<p>
+  <strong>
+    <%= _("Are you sure the information you’re after isn’t available " \
+       "elsewhere?") %>
+  </strong>
+  <%= _("Check the authority’s disclosure logs, their responses on " \
+        "{{site_name}} or conduct a quick Internet search.",
+        site_name: site_name) %>
+<p>
```
stevenday: These `<p>` tags are all being opened but not closed. I think perhaps there should be more of them too, if the newlines in the text is anything to go by?

PR #399, opened 2017-05-19
[#3708] Add a new contact us section for asking for help writing a request

lizconlan:
With the new section, the page will look like:

screen shot 2017-05-19 at 16 14 30

And with the section open:

contact us - whatdotheyknow 20170519

A bit of a non-ideal solution as now we have 2 copies of the form on the page but I can't think of how to fix that :(

Fixes mysociety/alaveteli#3708

### DIFF
--- a/lib/views/help/_contact_form.html.erb
+++ b/lib/views/help/_contact_form.html.erb
@@ -1,5 +1,5 @@
 <%= form_for :contact,
-             :url => help_contact_path + '#wdtk',
+             :url => help_contact_path + '#' +  form_id,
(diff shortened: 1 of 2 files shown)


lizconlan:
I think this is better behaved now - it's able to track which form section was open and reselect the correct one. It's still not idea if you've started filling in the form then choose to log in but I think that's a different (and pre-existing) issue

lizconlan:
> I've made an extra commit to add inline validations since we use them on some fields.

Nice! Thanks :cake:

PR #405, opened 2017-06-13
Tweak no-js version of logged in menu

crowbot:
Correct text colour, and hover/active/focus properties. Fixes https://github.com/mysociety/alaveteli-professional/issues/162. Since we've extended the main menu, the alignment issue has been made moot as we've dropped down the logged in menu in the non-js version.


stevenday:
Big screens see this:
screen shot 2017-06-19 at 15 33 25

Whereas smaller screens (but not small enough to trigger the mobile view) see this:
screen shot 2017-06-19 at 15 35 26

garethrees:
I think this is good to go with https://github.com/mysociety/alaveteli/pull/4186

PR #430, opened 2017-10-20
[pro#235] Missing cautionary text

lizconlan:
Copy the sidebar helper text about your message being public (with a modification to account for embargoes) to the pro sidebar template

screen shot 2017-10-20 at 14 06 10

Connects to mysociety/alaveteli-professional#235


garethrees:
Already added in aaf80752

Issue #439, opened 2017-11-16
Pro logo pixalated on retina screens


gbp:
![screenshot-2017-11-16-wdtk-pro](https://user-images.githubusercontent.com/5426/32905457-edb95152-caf1-11e7-9525-093afb4559a4.png)
![screenshot-2017-11-16-wdtk](https://user-images.githubusercontent.com/5426/32905458-ee0e6f84-caf1-11e7-9391-a3f1ca3adad4.png)

garethrees:
This got merged in https://github.com/mysociety/whatdotheyknow-theme/commit/42b20b2c97d8a2182a6faa438160571b43bd8e95

PR #440, opened 2017-11-23
[#439] retina logos

wrightmartin:
Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/439

- Adds new versions of logo files
- Adds SVG versions of logo files


wrightmartin:
**A note on filesizes**

The SVG versions of the logo are bigger files than the png ones, before gzipping by the server.

Post gzip file sizes are much better

screen shot 2017-11-22 at 17 09 28

app/assets/stylesheets/responsive/custom.scss
```diff
@@ -626,35 +626,35 @@ a.link_button_green_large {
...
 .site-title__logo {
   @extend .image-replacement;
-  width: 159px;
+  width: 165px;
   height: 18px;
-  background-size: 159px 18px;
+  background-size: 165px 18px;
   background-image: image-url("wdtk-logo-white.png");
+  // https://css-tricks.com/a-complete-guide-to-svg-fallbacks/
+  background-image: image-url("/assets/wdtk-logo-white.svg"),
```
garethrees: Looks like an incorrect URL – shouldn't need the `/assets/` prefix?

wrightmartin: oh - no it's wrong for another reason (I think)

Correct me if I'm wrong but `image-url()` doesn't appear to work with SVGs so it should be `url()`, that's why the `/assets/` prefix is there.

garethrees: Nerp – don't think so.

The asset path prefix is configurable, hence the helper.

I _think_ it should just be: `image-url("wdtk-logo-white.svg")` – I don't think the helper really cares about the file extension.

http://guides.rubyonrails.org/v4.2.10/asset_pipeline.html#css-and-sass

wrightmartin: Yes, that worked, when I tried it before it didn't, which probably means I had a typo in there somewhere. Fixed and pushed

PR #445, opened 2017-12-06
[pro#137] Hide donation buttons from pros

gbp:
Closes https://github.com/mysociety/alaveteli-professional/issues/137


lib/views/request/_sidebar.html.erb
```diff
@@ -1,6 +1,5 @@
 <div id="right_column" class="sidebar right_column" role="complementary">
-  <% unless flash[:request_sent] || \
-            (feature_enabled?(:alaveteli_pro) && current_user && current_user.is_pro?) %>
+  <% if show_donation_button? && flash[:request_sent] %>
```
garethrees: This should be a `!flash[:request_sent]`. We show it _unless_ `flash[:request_sent]` is set, because `flash[:request_sent]` makes a big fat "plz donate" partial appear, so we don't want to ask twice.

PR #449, opened 2018-01-08
add special logo for 10th birthday

wrightmartin:
screen shot 2018-01-08 at 09 53 48


lizconlan:
(have added the banner text in the same PR for convenience as they belong together!)

PR #460, opened 2018-02-09
[pro#356] Style details summaries like links

garethrees:
Seems like a sensible default given they're clickable.

Initially required to make them look nice when displaying public body
notes in the batch builder.

**Without this change:**

screen shot 2018-02-09 at 18 09 13

**With this change:**

screen shot 2018-02-09 at 18 07 56

See https://github.com/mysociety/alaveteli-professional/issues/356.

Counterpart to https://github.com/mysociety/alaveteli/pull/4518 – can be merged independently though.

### DIFF
--- a/app/assets/stylesheets/responsive/custom.scss
+++ b/app/assets/stylesheets/responsive/custom.scss
@@ -79,6 +79,12 @@ a:visited {
       color: $color_black;
     }
   }
+
+  details {
+    summary {
+      color: $color_pro_blue;
+    }
+  }
 }
 
 


app/assets/stylesheets/responsive/custom.scss
```diff
@@ -79,6 +79,12 @@ a:visited {
       color: $color_black;
     }
   }
+
+  details {
+    summary {
+      color: $color_pro_blue;
```
wrightmartin: You might also want to add `cursor: pointer;` here, as at least one browser I tested didn't do this by default

Issue #418, opened 2017-08-30
Authority page follower count border broken


garethrees:
Border extends below the bottom of the button:

![screen shot 2017-08-30 at 09 03 38](https://user-images.githubusercontent.com/282788/29861772-4f7810d6-8d62-11e7-9416-46ac07b85d28.png)

PR #465, opened 2018-05-03
Issues/418 fix follower count

wrightmartin:
Fixes the alignment issue described in #418 and abtracts the code into a mixin to cut down on reuse


wrightmartin:
I just realised from the commit history you can't see what the fix is:

Before when this code was duplicated across the stylesheet in one of the copies there was a 'font-size: 1em;' declaration https://github.com/mysociety/whatdotheyknow-theme/blob/master/app/assets/stylesheets/responsive/custom.scss#L976

This has been removed as part of turning it into a mixin.

PR #468, opened 2018-05-15
Remove mySociety header bar and logo

zarino:
For mysociety/research#123 – we’re dropping the mySociety logo from the headers of our end-user "tool" / "service" websites, to try and introduce some consistency.

screen shot 2018-05-15 at 16 53 21

Barely seems worth the screenshot, but here it is – no black bar!


garethrees:
👍

Issue #493, opened 2018-08-12
New house rules to cover abuse of the site


RichardTaylor:
The current house rules:

https://www.whatdotheyknow.com/help/house_rules

don't explicitly forbid creating new accounts to evade a ban, or to evade a rate-limiting cap. 

Multiple accounts are sometimes legitimate eg. a whistleblower who already has an account wanting to make a request anonymously. 

We probably don't want to expand the rules to cover every possible eventuality, there's some benefit in keeping them concise and readable. Something to consider.

RichardTaylor:
We've just had a user on WhatDoTheyKnow re-submit a request which administrators had hidden. 

A new rule, or rule, could cover this too; something like:

>Do not seek to evade the actions of site administrators by, for example, creating new accounts to evade a ban, or a rate-limiting cap, or by republishing material which you know has been removed by moderators.

garethrees:
@RichardTaylor if you're happy with the above wording we can get this updated.

garethrees:
I've made a PR with the above wording. We'll get this merged as-is unless we hear anything before Friday, but we can always change it later.

RichardTaylor:
I'm happy with using the words I proposed.

PR #504, opened 2018-09-19
[#493] Add House Rule for ban evasion

garethrees:
Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/493.


garethrees:
Richard happy with wording, so just going to merge as its minor text change https://github.com/mysociety/whatdotheyknow-theme/issues/493#issuecomment-423745268

Issue #500, opened 2018-09-06
Update privacy policy to extend data retention policy for ICO


garethrees:
https://github.com/mysociety/whatdotheyknow-theme/blob/master/lib/views/help/privacy.html.erb#L524-L530

Extend deadline to four weeks, and allow exceptional circumstances.

PR #505, opened 2018-09-19
[#500] Update Data Retention Policy

garethrees:
* Extend from 2 weeks to 4 weeks
* Add discretionary period after engagement with ICO commences

Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/500

### DIFF
--- a/lib/views/help/privacy.html.erb
+++ b/lib/views/help/privacy.html.erb
@@ -522,12 +522,18 @@
   </p>
   <p>
     For material that needs to be retained for a relatively short period e.g.
-    where we give the ICO two weeks to request the material from us before we
-    delete it, the material will remain hidden on WhatDoTheyKnow and will be
-    deleted at the end of the retention period. In the rare case where a longer
-    retention period is necessary, mySociety staff are responsible for moving
-    the material to non-web accessible storage on mySociety’s servers and
-    deleting it when the retention period is complete.
+    where we give the ICO four weeks to request the material from us before we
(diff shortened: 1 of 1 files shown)


garethrees:
@RichardTaylor I've had an initial stab at https://github.com/mysociety/whatdotheyknow-theme/issues/500. What do you think of the wording?

RichardTaylor:
Here's a proposal on the text from me:
>...where we give the ICO four weeks to request the material from us before we delete it, the material will remain hidden on WhatDoTheyKnow and will usually be deleted at the end of the retention period. In exceptional circumstances, such as where we are in correspondence with the ICO about the case, we may extend the retention period. 
>In any, rare, case where a retention period over four weeks is deemed necessary and it is considered the release of the material is would pose a significant risk of harm, mySociety staff are responsible for moving the material to non-web accessible storage on mySociety’s servers and deleting it when the retention period is complete.

I can't envisage us needing to invoke the last sentence once we've got an "in exceptional circumstances" line, but it is something we've considered doing, have planned for, and are ready to do if required, so we could mention it.

mdeuk:
> I can't envisage us needing to invoke the last sentence once we've got an "in exceptional circumstances" line, but it is something we've considered doing, have planned for, and are ready to do if required, so we could mention it.

I would agree with this stance - it is likely only to be used on the rarest of occasions; but, including it shows that we aim to be to uphold the highest standards of data stewardship, particularly in circumstances we deem to be especially unusual.

PR #512, opened 2018-10-04
Prevent spam terms being used in requests

garethrees:
Hack for WDTK-specific cases.

If an anonymous user tries to submit spam terms in a request, just
redirect to the home page.

If a signed in user tries to submit spam terms in a request, ban them,
sign them out and then inform them they’ve been banned by sending them
to their profile page where they’ll be shown the ban message.

We can generate tmp/spam_terms.txt manually for now.

### DIFF
--- a/lib/controller_patches.rb
+++ b/lib/controller_patches.rb
@@ -26,4 +26,56 @@ Rails.configuration.to_prepare do
 
   end
 
+
+  RequestController.class_eval do
+    before_action :check_spam_terms, only: [:new]
+
+    def check_spam_terms
+      return true unless params[:outgoing_message]
+      return true unless params[:outgoing_message][:body]
+
+      if spammer?(params[:outgoing_message][:body])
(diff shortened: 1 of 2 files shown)


garethrees:
In the rails console:

```ruby
msgs = OutgoingMessage.where('body ~ ?', 'scribd').select { |m| m.info_request.user.banned? }

File.open('/tmp/scribd-msgs.txt', 'w') do |file|
  msgs.each { |m| file.write(m.body) }
end
```

As the app user:

```sh
cat /tmp/scribd-msgs.txt | grep -Eo "scribd\.com\/doc\/[0-9]+\/" | sort | uniq > /tmp/scribd-urls.txt
cat /tmp/scribd-msgs.txt | grep -Eo "scribd\.com\/document\/[0-9]+\/" | sort | uniq >> /tmp/scribd-urls.txt

cat /tmp/scribd-urls.txt > /var/www/www.whatdotheyknow.com/shared/tmp/spam_terms.txt
```

PR #539, opened 2018-11-21
Don't show alternative contact options if there are no incoming messages

lizconlan:
Requires mysociety/alaveteli#4966


garethrees:
Approved but may need changes based on https://github.com/mysociety/alaveteli/pull/4966. Either way, don't need to re-review as its just theme reconciliation.

Issue #542, opened 2018-11-23
A/B tests on donate page


ajparsons:
Following on from [donation analysis](https://blogs.mysociety.org/internal/2018/11/23/donation-analysis/)  - this is the key ticket.

Basically we want to be able to run A/B tests on the location and content of the donate link on the WDTK requests page. @zarino says this is actually super easy and customisable through google analytics - so this is a ticket to a) ask someone to do this b) draw out any additional problems/info needed.

zarino:
Here’s an example of the JavaScript involved with a recent A/B test we did on the People’s Assembly homepage: https://github.com/mysociety/pombola/pull/2269/files

It pretty much boils down to:

```
window.cxVariation = cxApi.chooseVariation();
if(window.cxVariation === 1){
    // Change content on page
}
```

Then you set up your success criteria in the Content Experiment section of Google Analytics, and you’re away.

ajparsons:
For reference, I have created issue stubs for the first three experiments in the alaveteli-experiments repo

https://github.com/mysociety/alaveteli-experiments/issues/115
https://github.com/mysociety/alaveteli-experiments/issues/116
https://github.com/mysociety/alaveteli-experiments/issues/117

wrightmartin:
Can we add one to this for changing the design? I think the red box design is a big miss here - I'd guess it gets mistaken for an error message? I assume in the spirit of this ticket we would A/B test a design change?

wrightmartin:
Quick redesign of what the donate call _could_ look like

![desktop hd](https://user-images.githubusercontent.com/2292925/50159346-5e4af500-02ce-11e9-960b-85fddda6c4d6.png)

zarino:
@wrightmartin Is there a danger of it starting to look too much like a sidebar advert? The sort of thing we’ve all been trained to ignore ;-)

wrightmartin:
> @wrightmartin Is there a danger of it starting to look too much like a sidebar advert? The sort of thing we’ve all been trained to ignore ;-)

Yep - but it's okay, because we're testing that too https://github.com/mysociety/alaveteli-experiments/issues/115

garethrees:
So far:

> * https://github.com/mysociety/alaveteli-experiments/issues/115 – varying placement of banner

This made an improvement so we've made it permanent

> * https://github.com/mysociety/alaveteli-experiments/issues/116 – varying text of banner

`Donate £n Now` vs `Donate Now` doesn't seem to make much difference

[...]

PR #556, opened 2018-12-20
A/B test position of donate-cta on request page

zarino:
Moves the red donation banner from the sidebar to the end of the page, using JavaScript, if the current user is in Context Experiment group 1.

Also tracks clicks on the donation banner button, no matter where it’s located, so we can use that as the goal in our Content Experiment.

I’ve [set up the experiment](https://analytics.google.com/analytics/web/#/report/siteopt-experiments/a24614779w48006686p48388638/) but not started it yet. Guess we could start it now, it wouldn’t have any effect, but I figured we should get this PR merged and deployed first, just for tidiness ;-)

See mysociety/whatdotheyknow-theme#542 and (more specifically) mysociety/alaveteli-experiments#115.

screen shot 2018-12-20 at 15 27 09

### DIFF
--- a/app/assets/javascripts/tests.js
+++ b/app/assets/javascripts/tests.js
@@ -1,0 +1,33 @@
+(function($) {
+  $(function () {
+    window.cxVariation = cxApi.chooseVariation();
+    if(window.cxVariation === 1){
+      var $donate = $('.sidebar__donate-cta');
+      var $link = $('.sidebar__donate-cta__button', $donate);
+      var experimentHref = URI($link.attr('href')).removeSearch('utm_content').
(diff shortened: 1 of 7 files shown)


zarino:
@crowbot @garethrees It looks like I don’t have permission to create an Experiment at [whatdotheyknow.com analytics page](https://analytics.google.com/analytics/web/#/report/siteopt-experiments/a24614779w48006686p48388638/) – can you add that permission to my Google Analytics account?

gbp:
@garethrees @wrightmartin and I have gone through the frontend to verify the JS is working as expected. Could you review the small changes ruby helpers code.

PR #557, opened 2019-01-02
[#4985] Only show the followup form if the request is open to new responses

lizconlan:
Override the slightly simpler template, including the fix for showing part of the response form when new responses are not being accepted for the current request.

Connects to mysociety/alaveteli#4985
Requires mysociety/alaveteli#5033


lizconlan:
Good catch, thanks!

PR #570, opened 2019-04-29
[#564] Don't survey suspended users

### LINKED ISSUE #564: Users who are suspended should not be sent site surveys
There exists an unusual scenario where a user can create an account, make a request, and, then, be banned in short order.

If a user's account is banned prior to the survey being triggered (e.g. before two…ish weeks) it would perhaps be preferable if we did not send the "Can you help us improve WhatDoTheyKnow?" email, which is perhaps likely to receive a less favourable response than might be anticipated 🙂

Relates to #544

lizconlan:
## Relevant issue(s)?

Closes #564

## What does this do?

* Adds a `User#can_send_survey?` method to test whether the user is eligible to receive a survey alert
  * Adds `!user.active?` as a reason to not send the survey alert
* Delegates the user-based part of `RequestMailer.alert_survey`'s decision whether to send a survey alert to the new `User#can_send_survey?` method
* Uses `#deliver_now` instead of the deprecated `#deliver` method when sending survey alerts

## Why was this needed?

We want to stop sending survey alerts to suspended users (by checking the `User#active?` method)

### DIFF
--- a/spec/model_patches_spec.rb
+++ b/spec/model_patches_spec.rb
@@ -11,6 +11,53 @@ describe UserInfoRequestSentAlert, "when patched by the whatdotheyknow-theme" do
 
 end
 
+describe User, 'when patched by whatdotheyknow-theme' do
+
+  describe '#can_send_survey?' do
+    let(:user) { FactoryBot.create(:user) }
+    subject { user.can_send_survey? }
+
+    before do
+      allow(AlaveteliConfiguration).
+        to receive(:send_survey_mails).and_return(true)
+    end
+
+    context 'a survey has not been sent to an active user' do
+
+      before do
+        allow(user).to receive(:survey).
+          and_return(double('survey', :already_done? => false))
+      end
+
+      it { is_expected.to eq(true) }
+
+    end
+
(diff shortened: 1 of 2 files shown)


spec/model_patches_spec.rb
```diff
@@ -11,6 +11,46 @@
...
+    subject { user.can_send_survey? }
+
+    before do
+      allow(AlaveteliConfiguration).
+        to receive(:send_survey_mails).and_return(true)
+      allow(user).to receive(:survey).
+        and_return(double('survey', :already_done? => false))
+    end
+
+    it { is_expected.to eq(true) }
```
garethrees: Minor nitpick, but I'd make this its own context to make the "success" conditions more human readable:



```ruby
    #…
    subject { user.can_send_survey? }

    before do
      allow(AlaveteliConfiguration).
        to receive(:send_survey_mails).and_return(true)
    end

    context 'a survey has not been sent to an active user' do

      before do
        allow(user).to receive(:survey).
          and_return(double('survey', :already_done? => false))
      end

      it { is_expected.to eq(true) }
    end
```

PR #576, opened 2019-07-09
[core#4891] Revert quick fix

gbp:
This reverts #515 which re-enables the "Publish requests" button as https://github.com/mysociety/alaveteli/issues/4891 has been resolved.


gbp:
Merged into master f611f67

PR #577, opened 2019-07-11
Update credits

garethrees:
Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/578

* Add Matt and Martyn to the credits page.
* Remove CF from credits page
* Add staff & trustees

Screenshot 2019-07-29 at 16 48 09

### DIFF
--- a/lib/views/help/credits.html.erb
+++ b/lib/views/help/credits.html.erb
@@ -67,9 +67,11 @@
         <a href="https://www.whatdotheyknow.com/user/helen_cross_2">Helen
         Cross</a>,
         <a href="https://www.whatdotheyknow.com/user/john_cross">John Cross</a>,
-        <a href="https://www.whatdotheyknow.com/user/chris_fryer">Chris Fryer
-        </a>,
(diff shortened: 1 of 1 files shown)


lib/views/help/credits.html.erb
```diff
@@ -60,28 +60,31 @@
         maintain the database of public authorities and so much more. Thanks to
         <a href="https://www.whatdotheyknow.com/user/m_bimmler">Michael
         Bimmler</a>,
-        <a href="https://www.whatdotheyknow.com/user/kieran_casey_2">
-        Kieran Casey-McEvoy</a>,
-        <a
-        href="https://www.whatdotheyknow.com/user/gavin_chait/">Gavin Chait</a>,
+        <a href="https://www.whatdotheyknow.com/user/kieran_casey_2">Kieran
+          Casey-McEvoy</a>,
```
gbp: As you're after consistency, how about:

```suggestion
        Casey-McEvoy</a>,
```

garethrees: Sorry, yes, that should be as per suggestion. Fixed

PR #582, opened 2019-07-16
Add links to Companies House and ICO

### LINKED ISSUE #581: Include links to Companies House and ICO (DPR) registrations within public bodies
Similar to #486, we should include links, where appropriate, to Companies House and the ICO's data protection register.

We already tag relevant information within bodies using the tags 'ch' and 'dpr'; therefore, it may be possible to simply expand on the code within #491.

mdeuk:
This PR expands on #491 to include links to Companies House and the ICO's data protection register. It's conceivable that for some bodies we may only use the DPR, however, for others, we may end up using both.

Some examples to test with:
- [DfT OLR Holdings Limited](https://www.whatdotheyknow.com/body/dohl); has 'ch' tag
- [Glasgow City Council](https://www.whatdotheyknow.com/body/gcc); has 'dpr' tag
- [Network Rail Limited](https://www.whatdotheyknow.com/body/network_rail); has 'ch' and 'dpr' tags

This is the first time I've wrangled with the Alaveteli codebase, so, _please be gentle_ 😊

Closes #581

### DIFF
--- a/lib/views/public_body/_more_info.html.erb
+++ b/lib/views/public_body/_more_info.html.erb
@@ -22,6 +22,18 @@
                     "http://beta.charitycommission.gov.uk/charity-details/?regid=#{ tag_value }&subid=0" %><br>
     <% end %>
   <% end %>
+
+<% if public_body.has_tag?('ch') %>
+  <% public_body.get_tag_values('ch').each do |tag_value| %>
+      <%= link_to _('Company registration'),
+                    "https://beta.companieshouse.gov.uk/company/#{ tag_value }" %><br>
+  <% end %>
+
+<% if public_body.has_tag?('dpr') %>
+  <% public_body.get_tag_values('dpr').each do |tag_value| %>
+      <%= link_to _('Data Protection registration'),
+                    "https://ico.org.uk/ESDWebPages/Entry/#{ tag_value }" %><br>
+  <% end %>
 <% end %>
 
 <%= link_to _('View FOI email address'), view_public_body_email_path(public_body.url_name) %><br>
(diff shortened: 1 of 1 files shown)


RichardTaylor:
Fantastic. 

A couple of thoughts:

* We should list these new special tags on the authority admin page alongside the others to prompt admins to use them and to act a reminder of their format.  
* Is the "CH:" etc. case sensitive and should it be?  I suspect admins will have used various capitalisation eg. I see https://www.whatdotheyknow.com/body?tag=CH%3ASC356338 

We have a more general issue with tags being case sensitive and admins sometimes introducing something inconsistent eg. at the time of writing we have 8 public bodies tagged  

https://www.whatdotheyknow.com/body?tag=Scotland
and 540 tagged
https://www.whatdotheyknow.com/body?tag=scotland

Should we enforce lower case on tags?

mdeuk:
> * We should list these new special tags on the authority admin page alongside the others to prompt admins to use them and to act a reminder of their format.

As far as I can see, the configuration for this is held upstream at [/mysociety/alaveteli/blob/master/app/views/admin_public_body/_form.html.erb](https://github.com/mysociety/alaveteli/blob/master/app/views/admin_public_body/_form.html.erb), whereas, this change only affects a setting that is specific to WDTK. I'll let more learned minds comment on whether or not we can fill out wdtk-theme specific content there.

> * Is the "CH:" etc. case sensitive and should it be?  I suspect admins will have used various capitalisation eg. I see https://www.whatdotheyknow.com/body?tag=CH%3ASC356338

In this instance, convention indicates that lowercase should be used (as is already the case with the 'charity' tag, and, our categorisation tags - e.g. using the tag 'LLP' will produce a different result to 'llp').

gbp:
Superseded by #614

PR #584, opened 2019-07-24
[pro#587] Override the pro pricing controller to use markup in pro_site_name

lizconlan:
Requires mysociety/alaveteli/pull/5294
Related to mysociety/alaveteli-professional#587

Reinstates  the **WhatDoTheyKnow<sup>Pro</sup>** formatting on the pricing page otherwise the translatable template will render it as "**WhatDoTheyKnow Pro**"

### DIFF
--- a/lib/controller_patches.rb
+++ b/lib/controller_patches.rb
@@ -106,4 +106,13 @@ Rails.configuration.to_prepare do
     end
   end
 
+  AlaveteliPro::PlansController.class_eval do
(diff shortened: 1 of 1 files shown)


garethrees:
I honestly think we should just drop the superscript. Its fine in the logo but _everywhere_ else its a giant pain.

lizconlan:
Fair enough - closing this!

PR #591, opened 2019-07-31
Second phase of request page donation button A/B testing

zarino:
* Move the request page donation banner out of the sidebar, to its new home at the bottom of the correspondence list (replacing the AskTheEU banner that was there previously).
* Set up Google Optimize on the request pages, and have it pick one of three wording variations for the donation button.
* Trigger a Google Analytics event when the donation button is clicked.

Meanwhile, a [Google Optimize](https://optimize.google.com/optimize/home) experiment has been set up (as per https://github.com/mysociety/research/issues/194) with three variants (0, 1, and 2) and looking out for an "objective" of a Google Analytics event with a `category` of `Request page donate button`.

We’ll need to remember to "Start" the Optimize experiment, once we go ahead with all this.

Part of https://github.com/mysociety/alaveteli-experiments/issues/117.

PR #592, opened 2019-08-07
Fix Google Optimize loading on request pages

zarino:
Fix for two mistakes in 636b881 (original PR #591):

1. Makes sure that `dataLayer` is defined at pageload.
2. Avoids calling `ga('send', 'pageview')` twice.

Also, the `content_for :ga_pageview` block is now defined *before* the `render partial => 'ga_events'` block so our `ga` `create/require/send` calls are in the [same order that the docs recommend](https://support.google.com/optimize/answer/6262084). I don‘t think it makes a difference whether the pageview is sent before or after the Optimize ID is included, but at least this way it matches the docs.

@garethrees Once you merge and deploy this, I _think_ the Optimize experiment should work.

Related:

* https://github.com/mysociety/whatdotheyknow-theme/pull/591
* https://github.com/mysociety/research/issues/194

### DIFF
--- a/lib/views/request/show.html.erb
+++ b/lib/views/request/show.html.erb
@@ -6,6 +6,11 @@
 
 <% if flash[:request_sent] %><%= render :partial => 'request_sent' %><% end %>
 
+<% content_for :ga_pageview do %>
+  window.dataLayer = window.dataLayer || [];
+  ga('require', 'GTM-53XVLCX');
+<% end %>
+
 <%= render :partial => 'ga_events' %>
 
 <%= render :partial => 'hidden_request_messages' %>
@@ -109,11 +114,6 @@
 </div>
 
 <%- if @sidebar %><%= render :partial => @sidebar_template %><% end %>
-
-<%= content_for :ga_pageview do %>
-  ga('require', 'GTM-53XVLCX');
-  ga('send', 'pageview');
-<% end %>
 
 <%= content_for :javascript do %>
   <%= javascript_include_tag 'request-attachments.js' %>
--- a/app/assets/javascripts/tests.js
+++ b/app/assets/javascripts/tests.js
@@ -1,6 +1,3 @@
-// https://support.google.com/optimize/answer/9059383
-function gtag() { dataLayer.push(arguments); };
-
 function donateButtonTextVariant(variant) {
   if (variant === '0') {
       $(document).ready(function(){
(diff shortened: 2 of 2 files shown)


lib/views/request/show.html.erb
```diff
@@ -6,6 +6,11 @@
 
 <% if flash[:request_sent] %><%= render :partial => 'request_sent' %><% end %>
 
+<%= content_for :ga_pageview do %>
+    window.dataLayer = window.dataLayer || [];
+    ga('require', 'GTM-53XVLCX');
+<% end %>
```
garethrees: This is still going to render to the same place – https://github.com/mysociety/alaveteli/blob/develop/app/views/layouts/default.html.erb#L92

I wonder if the problem here is that `<%=` is rendering – usually you'd use `<%` so that the erb isn't rendered at this point. https://api.rubyonrails.org/v5.0.7/classes/ActionView/Helpers/CaptureHelper.html#method-i-content_for

zarino: Ah, I just used `<%=` because I was copying the `<%= content_for :javascript do %>` block immediately below. I’ve now tried it with and without the `=` and it doesn’t seem to make a difference to what’s output in the HTML.

BTW, I’m not able to verify your comment that "this is still going to render in the same place". Trying it out locally, when I have the following near the top of `request/show.html.erb`:

```erb
<% content_for :ga_pageview do %>
    window.dataLayer = window.dataLayer || [];
    ga('require', 'GTM-53XVLCX');
<% end %>
<%= render :partial => 'ga_events' %>
```

I get HTML output like this at the end of the request page:

```html
    ga('create', 'UA-24614779-1', 'auto');
            window.dataLayer = window.dataLayer || [];
    ga('require', 'GTM-53XVLCX');
    ga('send', 'pageview', '/request/the_cost_of_boring_two');
```

(Note, `pageview` at the end.)

Whereas, when I render the `ga_events` partial first, like so:

```erb
<%= render :partial => 'ga_events' %>
<% content_for :ga_pageview do %>
    window.dataLayer = window.dataLayer || [];
    ga('require', 'GTM-53XVLCX');
<% end %>
```

I get:

```html
    ga('create', 'UA-24614779-1', 'auto');
            ga('send', 'pageview', '/request/the_cost_of_boring_two');
    window.dataLayer = window.dataLayer || [];
    ga('require', 'GTM-53XVLCX');
```

This makes sense, since [the docs you linked to](https://api.rubyonrails.org/v5.0.7/classes/ActionView/Helpers/CaptureHelper.html#method-i-content_for) confirm that:

> `#content_for` concatenates (default) the blocks it is given for a particular identifier in order.

So the order in which `<%= render :partial => 'ga_events' %>` (and the `<% content_for :ga_pageview do %>` it contains) and our newly added `<% content_for :ga_pageview do %>` are run makes a difference to the order of the lines in the output HTML.

zarino: PS, I’ve added a fixup commit that swaps the `<%=` for a `<%` – even though it weirdly seems to make no difference, `<% content_for` is clearly the right way to do it.

zarino:
@liz I’ve popped the two bits of `dataLayer`-related code into the same conditional block – bit tidier.

Although, from a purist point of view, I _did_ wonder whether, rather than checking for the existence of google analytics, we should really be checking for the things the code actually uses, like:

```js
if ( typeof dataLayer === "object" && typeof gtag === "undefined" ) {
  function gtag() { dataLayer.push(arguments); };
}

if ( typeof gtag !== "undefined" ) {
  gtag('event', 'optimize.callback', {
    …
  });
}
```

Maybe too fussy though?

zarino:
> Although, from a purist point of view, I did wonder whether, rather than checking for the existence of google analytics, we should really be checking for the things the code actually uses, like:

I’ve made that change now. Feels much better.

@lizconlan I think this is ready for merge now.

PR #605, opened 2019-08-21
[Rails 5.1] Stop using alias_method_chain

lizconlan:
## Relevant issue(s)

Required by mysociety/alaveteli#3970

## What does this do?

~~Overrides `RequestMailer.alert_new_response_reminders` using `alias_method` (and repeating the original code/adding the extra line manually) rather than `alias_method_chain`~~

Patches RequestMailer using `include`/`prepend` with survey methods without using the old `alias_method_chain` and without duplicating code from Alaveteli core.

## Why was this needed?

`alias_method_chain` (deprecated in Rails 5.0) has been removed in Rails 5.1

### DIFF
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -91,8 +91,7 @@ Rails.configuration.to_prepare do
       end
     end
 
-    # Add survey methods to RequestMailer
-    RequestMailer.class_eval do
+    module SurveyMethods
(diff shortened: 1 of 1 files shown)


gbp:
@garethrees mind taking a quick look at this to make sure I haven't done anything stupid. This should be able to be merged as is and useable with current develop branch/Rails 5

PR #619, opened 2019-11-05
[#587] Add referral parameters in links to mySociety

### LINKED ISSUE #587: Some donation links do not have referral info
At https://www.whatdotheyknow.com/help/volunteers and https://www.whatdotheyknow.com/cy/help/volunteers the 'Donate here' links point to www.mysociety.org/donate.

I've noticed other donate buttons around the site contain links that identify the referral came from WDTK and when the user arrives at www.mysociety.org/donate they receive a targeted page with a header of Help keep WhatDoTheyKnow running which is probably more impactful towards WDTK users.

Presumably this affects tracking about which teams/sites donations originated from etc.

Could these links be changed?

andrewying:
Fixes #587

### DIFF
--- a/lib/views/alaveteli_pro/pages/_legal.html.erb
+++ b/lib/views/alaveteli_pro/pages/_legal.html.erb
@@ -188,7 +188,7 @@
   </h2>
   <p>
     You can read mySociety’s general privacy policy here:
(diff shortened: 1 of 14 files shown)


gbp:
Small changes made and merged 38fb25f

PR #627, opened 2019-11-14
Update credits page

andrewying:
Update credits page for transparency


gbp:
Split the commit into two and merged 7c30246

PR #634, opened 2019-12-06
Changes for WDTK pro landing pages

wrightmartin:
Changes for https://github.com/mysociety/alaveteli/pull/5493


garethrees:
@wrightmartin screenshot in the PR description would be good just to check what the expected differences are when we review this.

wrightmartin:
image

garethrees:
Going to close this as after a bit of tweaking of https://github.com/mysociety/alaveteli/pull/5493 and adding https://github.com/mysociety/alaveteli/pull/5498 we don't need the overrides anymore 🙌

PR #636, opened 2019-12-12
Update credits page

andrewying:
To add Matt Lewin as our volunteer


garethrees:
Thanks! Merged as faf1617987eb06316e2e832bda41532f5166861e (dropped the `target="_blank"`).

PR #661, opened 2020-03-23
Add Public Body name to requires admin email

gbp:
This will make support easier, especially for batch requests where almost identical emails can be received.


gbp:
Added to core https://github.com/mysociety/alaveteli/pull/5594

PR #663, opened 2020-03-26
Update company name to Society Works Ltd

garethrees:
We're just changing the commercial company name – we're not moving away
from "mySociety" in general.

Same company number https://beta.companieshouse.gov.uk/company/05798215.

Part of https://github.com/mysociety/orgsites/issues/1122


garethrees:
We've just noticed we refer a lot to "mySociety" throughout that file, e.g. 

https://github.com/mysociety/whatdotheyknow-theme/blob/1a1901c8718b0700368c6409112870f007c3cc71/lib/views/alaveteli_pro/pages/_legal.html.erb#L90

https://github.com/mysociety/whatdotheyknow-theme/blob/1a1901c8718b0700368c6409112870f007c3cc71/lib/views/alaveteli_pro/pages/_legal.html.erb#L190

It's not clear whether all that should also be "Society Works" – which gets confusing as most of the site references "built by mySociety".

abimysoc:
The terms on this page are part of a commercial agreement between the customer and SocietyWorks Ltd which sells the Pro service. Who built it and how it's referred to elsewhere aren't really relevant to that.

IMO the easiest solution is to dodge the problem by putting

>  The WhatDoTheyKnow Pro service (“WhatDoTheyKnow Pro”, the “Service”) is
    operated by SocietyWorks Ltd, company no. 5798215, of 483 Green Lanes,
    London N13 4BS (“we” or “us).

and then changing all references to mySociety to “we” “us” “our” etc as grammatically appropriate in context.

(also, there's a ” quotation mark missing after the word “us” there)

garethrees:
> …changing all references to mySociety to “we” “us” “our” etc as grammatically appropriate in context.

See what you think of c2eef91.

There's one outstanding question, which is what to do with 

https://github.com/mysociety/whatdotheyknow-theme/blob/1a1901c8718b0700368c6409112870f007c3cc71/lib/views/alaveteli_pro/pages/_legal.html.erb#L190

"our" is the obvious choice, but arguably it's a little weird that we then link out to mysociety.org. I assume SocietyWorks will get its own privacy policy that we can link to in the future, but for now, what to do? I don't have a strong preference.

> (also, there's a ” quotation mark missing after the word “us” there)

Fixed in a002c7e

garethrees:
Deployed https://www.whatdotheyknow.com/pro/pages/legal

PR #664, opened 2020-03-26
Update credits to remove a volunteer

MattK1234:
This is to remove a volunteer who no longer wishes to be included in the credits. The WDTK admin team have received a GDPR RtE request from them.

Support mail thread: https://mail.google.com/mail/u/1/#inbox/FMfcgxwHMPpmxhgSzjDqZXbgGHHVLvKp

Note: I think I've done this correctly, but it is my first time playing with this, apologies if it's incorrect! :-)


garethrees:
This seems to be missing the branch from your fork of the repo:

Screenshot 2020-04-02 at 12 32 49

Weird, since we still have access to the commit here. I've applied this as 062d20b, which preserves your authorship :).

Issue #670, opened 2020-03-31
Add Severnoaks DC autoresponder as an invalid reply address


MattK1234:
`autoresponder@sevenoaks.gov.uk` doesn't accept replies. This caused a user confusion today.

Seems it needs to be added to `ReplyToAddressValidator.invalid_reply_addresses` at https://github.com/mysociety/whatdotheyknow-theme/blob/master/lib/model_patches.rb

I should be able to get a PR in for this in the next couple of days.

MattK1234:
I also propose adding a page to the internal wiki about this.

garethrees:
Fixed by https://github.com/mysociety/whatdotheyknow-theme/pull/673

MattK1234:
I've added a page to the internal Wiki at https://wdtkwiki.mysociety.org/wiki/Invalid_Reply_Addresses for future reference.

PR #673, opened 2020-04-02
Add Sevenoaks responder as invalid reply address

MattK1234:
Adds `autoresponder@sevenoaks.gov.uk` as an invalid reply 
address meaning users will not be able to send follow-ups or internal 
review requests to it.

This caused confusion for a user who contacted us because they had received a delivery failure message after replying to the above.

Linked to #670


garethrees:
Perfect! Thanks so much!

Issue #672, opened 2020-04-02
Temporary late calculation for Scottish authorities


garethrees:
Looks like Scotland have extended the deadlines to 60 days (normal) / 100 days (extension) if I'm reading this correctly http://www.itspublicknowledge.info/home/News/20200401.aspx

I think we can fairly easily handle this in WDTK by adding a custom late calculator based on the `scotland` tag:

```diff
diff --git a/lib/model_patches.rb b/lib/model_patches.rb
index 7344b39..6301726 100644
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -54,7 +54,9 @@ Rails.configuration.to_prepare do
 
         def late_calculator
           @late_calculator ||=
-            if public_body.has_tag?('school')
+            if public_body.has_tag?('scotland')
+              ScotlandLateCalculator.new
+            elsif public_body.has_tag?('school')
               SchoolLateCalculator.new
             else
               orig_late_calculator
diff --git a/lib/scotland_late_calculator.rb b/lib/scotland_late_calculator.rb
new file mode 100644
index 0000000..8074f6b
--- /dev/null
+++ b/lib/scotland_late_calculator.rb
@@ -0,0 +1,14 @@
+# -*- encoding : utf-8 -*-
+class ScotlandLateCalculator
+  def self.description
+    %q(Scotland has extended deadlines during COVID-19)
+  end
+
+  def reply_late_after_days
+    60
+  end
+
+  def reply_very_late_after_days
+    100
+  end
+end

```

I had a _very quick_ look at some (~5) Scottish and they all had the `scotland` tag. I expect we're missing some, but we can deal with any missing if they get raised to user support.

garethrees:
> the Scottish Government’s proposed law would extend the initial deadline for public bodies to respond to FoI requests from 20 to 60 working days, with a possible further extension of 40 working days.
>
> https://theferret.scot/coronavirus-bill-scottish-information-commissioner/

So yes, initial of 60, +40 = 100 for very late.

MattK1234:
> I had a very quick look at some (~5) Scottish and they all had the scotland tag. I expect we're missing some, but we can deal with any missing if they get raised to user support.

We've also got 8 tagged `Scotland`. I'm going to change them now to `scotland` to make life easier.

MattK1234:
> We've also got 8 tagged `Scotland`. I'm going to change them now to `scotland` to make life easier.

Done!

[...]

PR #675, opened 2020-04-02
Add ScotlandLateCalculator

garethrees:
Scotland has extended deadlines during COVID-19. This applies a custom
LateCalculator to authorities tagged `scotland` to reflect the new
timelines.

Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/672.

**New late calculation** for authority tagged `scotland`:

Screenshot 2020-04-02 at 11 49 08

Remove the late calc – still get the same due date for requests made during covid-19.

Screenshot 2020-04-02 at 11 51 26

Not sure what situations this might get recalculated in, but I imagine we'll be _mostly_ correct. Given its a temporary measure, I think we can live with a few deviations.

### DIFF
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -54,7 +54,9 @@ Rails.configuration.to_prepare do
 
         def late_calculator
           @late_calculator ||=
-            if public_body.has_tag?('school')
+            if public_body.has_tag?('scotland')
+              ScotlandLateCalculator.new
+            elsif public_body.has_tag?('school')
               SchoolLateCalculator.new
             else
               orig_late_calculator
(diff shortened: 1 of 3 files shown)


lib/model_patches.rb
```diff
@@ -54,7 +54,9 @@ def email_subject_request(opts = {})
 
         def late_calculator
           @late_calculator ||=
-            if public_body.has_tag?('school')
+            if public_body.has_tag?('scotland')
+              orig_late_calculator
```
gbp, line 58: Shouldn't this be:
```suggestion
              ScotlandLateCalculator.new
```

garethrees: Oh oops! added that to test it out and clearly to eager on the `git add .` 😊

PR #682, opened 2020-04-23
Add H&F In Touch email as an invalid reply address

MattK1234:
Adds `H&FInTouch@lbhf.gov.uk` to the list of invalid reply addresses. This should prevent replies etc being sent to it which bounce (see mysociety/alaveteli#3465)

We've had a few of these for this email address, hopefully this will be a workaround to prevent problems for users.


MattK1234:
I contemplated which email address I should add here. I settled for the `H&FInTouch@domain` as opposed to `H\&FInTouch@domain` as the former is the one that appears in the request event log. Do correct me if that's wrong!

garethrees:
Thanks! Will get this deployed in the next few days.

PR #695, opened 2020-06-23
Company name change

abimysoc:
I think this PR captures all references to the old company name, and replaces them with mySociety.


gbp:
Seems to be another reference https://github.com/mysociety/whatdotheyknow-theme/blob/master/lib/views/help/officers.html.erb#L14

abimysoc:
Dunno why that didn't come up in my search first time around :-/ Changed.

PR #706, opened 2020-08-04
Improve adding custom inclusion validation options

gbp:
We were relying on Rails to not change the order of validation callbacks
but in Rails 6 this has now changed. Now we find the relevant callback
using class and attribute before updating its options.


garethrees (approved):
Code seems sensible, but kinda hard to review this as it doesn't cite any Rails changelog / documentation / source code for me to really understand the change, or a before/after snippet for me to be able to easily test it out.

Given I get the gist of the change though, I'm happy if you are.

PR #707, opened 2020-08-16
Fix footer link "Run by volunteers"

frabcus:
Linking to "Want to help?" section was confusing, it is much clearer to link to the "Volunteers" section, and credits them properly.

I think it used to say "Run by volunteers... and you!" and the "and you!" linked to "Want to help?". I'm guessing it got stuck with that value while being updated.


garethrees:
Thanks! 👌

PR #721, opened 2020-10-08
[#3465] Add one new address to invalid list model_patches.rb

mdeuk:
## Relevant issue(s)
mysociety/alaveteli#3465

## What does this do?

This patch adds an email address for Oxfordshire County Council (as seen in request 684973 and 690752) to the invalid list.

## Why was this needed?

This is needed as a workaround for the issue described above which is affecting some WDTK users who attempt to reply to a specific mailbox that Oxfordshire County Council is issuing responses from.

## Implementation notes
N/A, this is a straightforward addition to existing code.

## Screenshots
N/A

## Notes to reviewer

This does the same as #720, sadly it arose just after the previous patch was committed. 🙃

It's worth noting that the email address quoted here has a capital letter in the domain name - this is consistent with how it is presented by the OCC email system in replies from their staff (e.g. in message [1653110](https://www.whatdotheyknow.com/admin/incoming_messages/1653110/edit)).

### DIFF
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -154,6 +154,7 @@ Rails.configuration.to_prepare do
     DoNotReply@dhsc.gov.uk
     OSCTFOI@homeoffice.gov.uk
     SOCGroup_Correspondence@homeoffice.gov.uk
+    FOI-E&E@Oxfordshire.gov.uk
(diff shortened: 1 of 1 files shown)


garethrees:
✨

Issue #722, opened 2020-10-08
Implement technical changes to reduce volumes of personal correspondence sent to key bodies


mdeuk:
Although the majority of our users utilise WDTK to make transparency requests to public bodies, there are occasions where users are utilising WDTK as a method of contact for personal correspondence, e.g. to request information about their own social security benefits, their driving licence, or their biometric residence permit.

We have already implemented changes under #554 which adds questions to the request form for the Home Office. This has been moderately successful, however there are still a good number of requests which have to be manually removed by WDTK administrators afterwards.

A reasonable mitigation may be to include, where possible, a contact address for the public body, so that the user is signposted elsewhere.

mdeuk:
Suggested bodies to look at:

Department for Work and Pensions `(dwp - 24)` 
HM Revenue and Customs `(hmrc - 592)`
Driver and Vehicle Licensing Agency `(dvla - 719)`
Driver and Vehicle Agency (Northern Ireland) `(dvani - 3300)`
HM Passport Office `(hm_passport_office - 590)`

We should also look at adding the email address for the Home Office `(home_office - 7)` into the existing codebase.

mdeuk:
> We should also look at adding the email address for the Home Office `(home_office - 7)` into the existing codebase.

Added public.enquiries@ email address in d88e588; I'll PR this in due course.

garethrees:
> Added public.enquiries@ email address in [d88e588](https://github.com/mysociety/whatdotheyknow-theme/commit/d88e588ac2969de9d5408dd072d214d97af3ca07); I'll PR this in due course.

Merged in 5f630c5.

mdeuk:
Of note, our colleagues over in Australia have a nice generic ["stop and think" speedbump](https://www.righttoknow.org.au/new/nsw_police_force) that they deploy. The technical implementation is in [/views/requests/new.html.erb](https://github.com/openaustralia/righttoknow/blob/ace85f8376e2a753f89daf4e223d62e3b2dce5a4/lib/views/request/new.html.erb#L233) and **_appears_** to be based on how many hidden requests a body has - nicely done!

@FOIMonkey has suggested reaching out to find out if this seems to be effective at reducing takedowns.

HelenWDTK:
This issue is being closed due to a lack of discussion or resolution for over 12 months. Should we decide to revisit this issue in the future, it can be reopened.

PR #723, opened 2020-10-08
Add home office email to public_body_questions.rb

mdeuk:
## Relevant issue(s)
#722 

## What does this do?
This patch adds the  Home Office public enquiries email to 'home_office_deny_response' within public_body_questions.rb

## Why was this needed?
When we present users with the refusal prompt contained in 'home_office_deny_response' we don't provide them with an explicit 'out', e.g. we don't tell them how to contact the public body themselves. This is perhaps leading to users bypassing this option and choosing to continue using WDTK for requests concerning their own circumstances (BRP / Visa / Entitlements etc).

## Implementation notes
Nothing specific - this adds to existing code.

## Screenshots
N/A

## Notes to reviewer

This is an interim hack - it does two things, it adds `<strong>` tags to the phrase "which makes all correspondence public" and adds a paragraph containing the Home Office public enquiries email.

I did ponder providing a link to the relevant section on the [HO website](https://www.gov.uk/government/organisations/home-office#org-contacts), but opted for this as a simpler signpost.

### DIFF
--- a/lib/public_body_questions.rb
+++ b/lib/public_body_questions.rb
@@ -14,11 +14,17 @@ Rails.configuration.to_prepare do
         you get a response. This has the benefit of highlighting difficulties
         communicating with the Home Office to MPs.
       </p>
+      <p>
+      You could also try contacting the Home Office by email at: <a
+      href="mailto:public.enquiries@homeoffice.gov.uk">public.enquiries@
+      homeoffice.gov.uk</a>.
+      </p>
 
       <p>
-        Misusing our service, which makes all correspondence public, won't help
-        you pursue your individual case as the Home Office will not enter into
-        correspondence about individual cases via our service.
+        Misusing our service, <strong>which makes all correspondence
+        public</strong>,won't help you pursue your individual case as the Home
+        Office will not enter into correspondence about individual cases via our
+        service.
       </p>
     HTML
   )


lib/public_body_questions.rb
```diff
@@ -14,11 +14,17 @@
...
+      href="mailto:public.enquiries@homeoffice.gov.uk">public.enquiries@
+      homeoffice.gov.uk</a>.
+      </p>
 
       <p>
-        Misusing our service, which makes all correspondence public, won't help
-        you pursue your individual case as the Home Office will not enter into
-        correspondence about individual cases via our service.
+        Misusing our service, <strong>which makes all correspondence
+        public</strong>,won't help you pursue your individual case as the Home
```
garethrees, line 25: Minor typo here:

```diff
- public</strong>,won't
+ public</strong>, won't
```

garethrees (commented):
I'm happy to merge this, but it did make me question whether this is an approach we want to start taking.

It certainly feels like the easiest route to reducing the amount of misuse, but it does come with the side effect of needing to keep these emails up to date (or, in the case that we don't, users getting bounces – maybe we don't mind, as long as they don't misuse WDTK?).

I guess the set of authorities where we feel the need to do this is small enough for it not to be a huge issue, and if we start needing to do more of this it'll become painful and we can think of a new approach at that point.

mdeuk:
> I'm happy to merge this, but it did make me question whether this is an approach we want to start taking.
> 
> It certainly feels like the easiest route to reducing the amount of misuse, but it does come with the side effect of needing to keep these emails up to date (or, in the case that we don't, users getting bounces – maybe we don't mind, as long as they don't misuse WDTK?).
> 
> I guess the set of authorities where we feel the need to do this is small enough for it not to be a huge issue, and if we start needing to do more of this it'll become painful and we can think of a new approach at that point.

It's a valid point - the alternative is that we could use the contacts page on the respective homepage of the public bodies that we choose to list; for the Home Office that'd work without issue since the email address is listed there anyway.

From a maintainability perspective, would that be more suitable? 😀

PR #732, opened 2020-11-13
Update links to charity/company databases in _more_info.html.erb

### LINKED ISSUE #711: Charity Commission links do not depreciate gracefully
The Charity Commission for England and Wales have [updated](https://charitycommission.blog.gov.uk/2020/09/03/the-online-register-of-charities-new-and-improving/) their searchable database, and in doing so, have setup redirects from the old systems (apps and beta). However, these links do not depreciate gracefully, as the item that the user has searched for (in our case, the charity number we were passing in a URL) is not displayed, meaning the search has to be manually re-run.

This affects our "charity registration" links in the sidebar (e.g. those introduced in #582 / #614); but also affects our ["about" page](https://github.com/mysociety/whatdotheyknow-theme/tree/master/lib/views/help/about.html.erb) , which has a link to the older database on "apps".

mdeuk:
## Relevant issue(s)
Fixes #711 

## What does this do?
This patch updates the URLs that we use for Charity Commission and Companies House databases in the _more_info.html.erb sidebar.

The URL format for Companies House is a straight find-and-replace of the domain name, replacing `beta.companieshouse.gov.uk` with `find-and-update.company-information.service.gov.uk`.

The URL for the Charities Commission is a more significant change - replacing the domain name and parameters.

## Why was this needed?
Both databases have moved from beta to production, and as such, the domain name and URL format have changed. This is more problematic for the Charity Commission as links from the old production and beta systems do not degrade gracefully.

## Implementation notes
This is an update to existing code - there are no specific tests that need to be conducted; however, to ensure that the correct charity link is being rendered, it may be helpful to check a few bodies.

To test the charity commission database:

1. Select the public body page for an appropriate [charity](https://www.whatdotheyknow.com/body?tag=charity), for example [Dudley and West Midlands Zoological Society Limited](https://www.whatdotheyknow.com/body/dudley_zoo) (an English charity), [North Ayrshirie Leisure Limited](https://www.whatdotheyknow.com/body/ka_leisure) (Scotland).
2. Ensure that the 'Charity Registration' link is being rendered in the sidebar
3. Check that the correct link is being registered. For charities with a registration number starting 'SC' this should be a link to OSCR; otherwise, it should be a link to the Charity Commission.

## Screenshots
N/A

## Notes to reviewer
N/A

### DIFF
--- a/lib/views/public_body/_more_info.html.erb
+++ b/lib/views/public_body/_more_info.html.erb
@@ -19,7 +19,7 @@
                     "https://www.oscr.org.uk/about-charities/search-the-register/charity-details?number=#{ tag_value }" %><br>
     <% else %>
       <%= link_to _('Charity registration'),
-                    "http://beta.charitycommission.gov.uk/charity-details/?regid=#{ tag_value }&subid=0" %><br>
+                    "https://register-of-charities.charitycommission.gov.uk/charity-search/-/results/page/1/delta/20/keywords/#{ tag_value }" %><br>
     <% end %>
   <% end %>
 <% end %>
--- a/lib/views/public_body/_more_info.html.erb
+++ b/lib/views/public_body/_more_info.html.erb
@@ -27,7 +27,7 @@
 <% if public_body.has_tag?('ch') %>
   <% public_body.get_tag_values('ch').each do |tag_value| %>
       <%= link_to _('Company registration'),
-                    "https://beta.companieshouse.gov.uk/company/#{ tag_value }" %><br>
+                    "https://find-and-update.company-information.service.gov.uk/company/#{ tag_value }" %><br>
   <% end %>
 <% end %>
 


mdeuk:
### Notes to reviewer (updated)
I've updated the original issue, https://github.com/mysociety/whatdotheyknow-theme/issues/711#issuecomment-728145412, with some background on the change proposed in 8f8ee82677f6bb4ddb2b88a092511c952511e48f, explaining the rationale behind this in light of some new information.

We may wish to consider whether or not this element of the proposed PR should be implemented - my gut instinct says 'yes', but it may be worthwhile considering whether or not sticking closer to our original is the better thing to do. Comments most welcome.

lib/views/public_body/_more_info.html.erb
```diff
@@ -19,7 +19,7 @@
                     "https://www.oscr.org.uk/about-charities/search-the-register/charity-details?number=#{ tag_value }" %><br>
     <% else %>
       <%= link_to _('Charity registration'),
-                    "http://beta.charitycommission.gov.uk/charity-details/?regid=#{ tag_value }&subid=0" %><br>
+                    "https://register-of-charities.charitycommission.gov.uk/charity-search/-/results/page/1/delta/20/keywords/#{ tag_value }" %><br>
```
garethrees, line 22: OOI, why not link [directly to the page for the charity](https://register-of-charities.charitycommission.gov.uk/charity-search/-/charity-details/507221), rather than the [search page](https://register-of-charities.charitycommission.gov.uk/charity-search/-/results/page/1/delta/20/keywords/507221)?

garethrees: Doh, sorry, missed the [comment above](https://github.com/mysociety/whatdotheyknow-theme/pull/732#issuecomment-728146916)

mdeuk: It's a reasonable question… unfortunately I've not had my coffee yet, so here's some unstructured thoughts:

If we link directly to the page for the charity itself, there is a chance we're not actually getting the right body.

In the example at https://github.com/mysociety/whatdotheyknow-theme/issues/711#issuecomment-728145412 there are multiple bodies that use the same charity number (which then has a sequential identifier applied to it). If we pass the charity number to the _direct_ page the system will not render any of these 'linked' bodies. This appears to be important, because it could be a 'linked' body that we list, rather than the parent.

There is a way to locate the body within the page displayed, but it is messy, since it requires multiple steps to be followed (click on "Linked Charities", then click the name of the charity, then click "Linked charity $number"). If we render the search page instead, the user gets a choice as to which one they want.

I'm agnostic on this - the solution evolved slightly since the patch was initially written; if there's an easier way that we think will be sufficient I'd be happy to go with that instead. 👍

PR #738, opened 2020-12-01
Add Castle Point BC invalid reply address

MattK1234:
## Relevant issue(s)
N/A - No issue raised for this one.

## What does this do?
This patch adds a new no-reply addresses for Castle Point Borough Council to model_patches.rb

## Why was this needed?
Correspondence being issued from public bodies from 'no-reply' email addresses which our users were unable to respond to. This particular address was located on the thread at https://www.whatdotheyknow.com/request/fraud_investigation_61 to Castle Point Borough Council.

## Implementation notes
N/A

## Screenshots
N/A

## Notes to reviewer
N/A


mdeuk:
LGTM 👍

Hopefully this can be deployed at thee same time as #731 and #734 - it'd be good to get all four addresses onto the system soon.

PR #744, opened 2020-12-10
Configure refusal references

garethrees:
Requires https://github.com/mysociety/alaveteli/pull/6015.

Adds refusals/exemptions from the Freedom of Information Act 2000 [1].

Fixes https://github.com/mysociety/alaveteli/issues/6002.

[1] https://foiwiki.com/foiwiki/index.php/Freedom_of_Information_Act_2000

### DIFF
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -46,6 +46,13 @@ Rails.configuration.to_prepare do
     end
   end
 
+  Legislation.refusals = {
+    foi: ['s 11', 's 12', 's 14', 's 21', 's 22', 's 22A', 's 23', 's 24',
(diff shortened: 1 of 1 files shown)


gbp (commented):
I'm not sure this is the best public interface to define these. Is there any reason why these are set on the Legislation class instead of the individual instances?

I would expect something like:

```ruby
foi = Legislation.find('foi') # or Legislation.default
foi.refusals = [...]
```

We would need to use class variables in `Legislation.all` so the instances aren't redefined every time its called.

If its a matter if merging to make progress then we can revisit this when we create a proper DB backed Legislation object.

garethrees:
https://github.com/mysociety/alaveteli/pull/6015 has been merged so this is ready for merge.

PR #755, opened 2021-01-25
[#754] Add version information to House Rules

### LINKED ISSUE #754: Add version information to House Rules
We should, for the purposes of transparency, provide some version information on the House Rules, to indicate when they were last changed, and what that change might have been.

For practicality purposes, I'd propose that we use a solution in a similar vein to that of the Privacy Notice (#700), piggybacking off our existing version control logs.

mdeuk:
## Relevant issue(s)
Fixes #754 

## What does this do?
This change adds a link back to the version control logs, allowing users to identify when changes have been made to the House Rules.

## Why was this needed?
We didn't have a specific link within the House Rules to allow users to see changes that had been made.

## Implementation notes
This implements a new paragraph with semantic tagging towards the bottom of /lib/views/help/house_rules.html.erb - it is based on the text already used in the privacy notice, with an updated URL and some minor tweaks.

## Screenshots
N/A

## Notes to reviewer
N/A

### DIFF
--- a/lib/views/help/house_rules.html.erb
+++ b/lib/views/help/house_rules.html.erb
@@ -83,5 +83,18 @@
     that your intentions are not malicious, we will contact you first so that we
     can give advice on how better to use the service.
   </p>
+  <h2 id="changes-to-house-rules">
+  Changes to our House Rules
+  <a href="#changes-to-house-rules">#</a>
+</h2>
+
+<p>
+  We keep these rules under review, and may make changes from time to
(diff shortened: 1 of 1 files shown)


garethrees:
Nice. We'll take a look at this when we get chance (likely won't be for a few days or more), but one thing I'm wondering is whether we can programmatically include this on each help page. Should be doable to render the "current template" path into the GitHub URL.

mdeuk:
I'll back this out in favour of the better change proposed in #761 👍

Issue #700, opened 2020-06-29
Privacy Notice should show version information


mdeuk:
We should, for transparency at a minimum, show some version information on the Privacy Notice - e.g. the date it was last revised or a version number.

If necessary, we could expand on this to include basic information about _what_ changed (e.g. we've updated our policy on x, or we've changed the name of the limited company) to aid understanding of how we seek to act in an open and transparent manner.

A good example of transparency is the ICO's own Privacy Notice which contains a [change log](https://ico.org.uk/global/privacy-notice/cha-nges-to-this-privacy-notice/) that can be referred to by the 'average user'.

mdeuk:
> 
> 
> The authoritative history is https://github.com/mysociety/whatdotheyknow-theme/commits/master/lib/views/help/privacy.html.erb (and the "version" is the SHA of the most recent commit).
> 
> Could link to that URL?

I think that's probably fine - an idea from the Github and gov.uk policies would be to include the date the policy change takes effect in the policy itself. However, we could probably back that off to the commit history if necessary.

garethrees:
> …include the date the policy change takes effect in the policy itself. However, we could probably back that off to the commit history if necessary.

Yeah, I'd rather not have to remember to keep dates up to date if we can avoid it.

PR #761, opened 2021-02-02
Help page version history

garethrees:
Adds a generic "changes" section to the bottom of most help pages.

Avoids having to write this individually each time we think we need it (#700, #754) and keeps things like the page anchors consistent.

Screenshot 2021-02-02 at 13 29 14


mdeuk (approved):
This looks fantastic, and very much keeps things in line with how a 'knowledgebase' type layout should work.

Cheers! 👍

gbp (approved):

PR #764, opened 2021-02-10
[#763] Add semantic tags to /lib/views/help/how.html.erb

### LINKED ISSUE #763: "How we run WhatDoTheyKnow" Help page does not have bookmark tags
The help page [/help/how](https://www.whatdotheyknow.com/help/how) does not have bookmarks that allow users to easily find a given part of the page.

It would be helpful to our users if this was included, since it would provide a more intuitive experience, particularly in a somewhat lengthy (by necessity) document.

mdeuk:
## Relevant issue(s)
Fixes #763 

## What does this do?
This adds and improves the semantic tagging used within /lib/views/help/how.html.erb

## Why was this needed?
The current presentation of the file makes it difficult for users (and WDTK administrators) to easily cite individual parts of the guidance.

## Implementation notes
Nothing specific to note - the changes are, for the most part, minor additions to existing tags (e.g. addition of 'id' tags to specific elements); however, a number of `<strong>` tags have also been replaced with headings in order to improve presentation.

## Screenshots
N/A

## Notes to reviewer
The commit at 2f65cd1 adds links to the "What happens" section - I _think_ this makes sense, but any feedback is welcome.

The commit at 8d3b0ea adds a reference back to the contact form. We might want to replace this with a more specific link if we improve the forms in the future.

### DIFF
--- a/lib/views/help/how.html.erb
+++ b/lib/views/help/how.html.erb
@@ -1,7 +1,7 @@
 <% @title = "How we run WhatDoTheyKnow" %>
 <%= render :partial => 'sidebar' %>
 <div id="left_column_flip" class="left_column_flip">
-  <h1 id="title"><%= @title %></h1>
+  <h1 id="title"><%= @title %> <a href="#title">#</a></h1>
   <p>
     WhatDoTheyKnow is administered by a small group of dedicated
     <a href="<%= help_credits_path(:anchor => 'volunteers') %>">volunteers</a>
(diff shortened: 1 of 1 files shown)


MattK1234:
This looks excellent! Thanks @mdeuk. From a user support point of view these tags make it a lot easier to point users towards certain parts of the page, saving admin time in the process.

gbp:
Nice work. Merged as 2e9cda41d486fddd362634b1b08453d82f032e59

PR #770, opened 2021-02-11
Reflect our new complaints process

crowbot:


lib/views/help/complaints.html.erb
```diff
@@ -35,14 +35,14 @@
         Having been through these two phases, should you remain unhappy with a
```
mdeuk, line 35: It might be a good idea for us to add anchors and headings  e.g. to clearly delineate stages of our process?

garethrees: -1 in this case.

I think top-level anchors per-section is generally fine, and this process should be understood as a whole.

If we wanted to improve this I'd add each stage as a subheading on the page.

PR #789, opened 2021-03-12
Update privacy text to reflect other reasons we might retain support mail

crowbot:


mdeuk (approved):
This looks good to me 👍

PR #813, opened 2021-05-17
[alaveteli#6236] Reposition refusal advice wizard

garethrees:
Fixes https://github.com/mysociety/alaveteli/issues/6236.

* Keep the high level /help/unhappy intro
* Add a line that explains that the user can use the refusal advice
  wizard if they’re unsure of what to do
* Remove left_column id/class to make page full-width. The wizard feels
  really cramped at a narrower width.

Screenshot 2021-05-17 at 17 33 40


garethrees:
Closing as in discussing https://github.com/mysociety/alaveteli/issues/6235, we decided that actually the _bottom_ of the page is a better place for this.

PR #817, opened 2021-05-18
Polish help unhappy

garethrees:
Fixes https://github.com/mysociety/alaveteli/issues/6236
Fixes https://github.com/mysociety/alaveteli/issues/6235
Requires https://github.com/mysociety/alaveteli/pull/6258 to show unfold arrows

Screenshot 2021-05-18 at 16 43 48

### DIFF
--- a/config/refusal_advice/actions.yml.erb
+++ b/config/refusal_advice/actions.yml.erb
@@ -33,5 +33,5 @@ foi:
     title: <%= _('Send a followup') %>
     body:
       plain: >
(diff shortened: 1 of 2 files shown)


garethrees:
@wrightmartin does this address most of the recent thinking in https://github.com/mysociety/alaveteli/issues/6236?

I don't want to spend loads of time making big changes if what we already have is Good Enough, but if there are minor tweaks we think we should make then happy to incorporate.

wrightmartin:
@garethrees yep, loads better

PR #822, opened 2021-05-27
Change final statement of requests wizard to work without suggested actions

wrightmartin:
Closes https://github.com/mysociety/alaveteli/issues/6244


garethrees (approved):
👌

PR #846, opened 2021-07-07
WiP Add Scottish legislations

gbp:
Adding refusal advice for FOISA was briefly discussed on standup. Would require us to add Scottish FOI and EIR legislations. This is an initial attempt. Still needs testing.


gbp:
When adding the `scotland` tag to an authority I am seeing these legislations being used by the app:

image

PR #882, opened 2021-09-08
[#881] Add citrix sharefile notification address to model_patches.rb

### LINKED ISSUE #881: Add Citrix ShareFile Notification address to model_patches.rb
Some public bodies, such as Aberdeen City Council, are using Citrix's ShareFile product to send large attachments.

This product generates [automated notifications](https://www.whatdotheyknow.com/admin/incoming_messages/1708598/edit), which although giving the appearance they are perhaps sent by a named person, do not accept replies, as they are sent from a generic no-reply email address.

We should add `mail-AT-sf-notifications-dot-com` to the list in model_patches.rb to prevent this.

mdeuk:
## Relevant issue(s)
Fixes #881 

## What does this do?
This patch adds the Citrix ShareFile notification email address to model_patches.rb.

## Why was this needed?
This is needed to prevent users from sending messages to a no-reply address (which is unroutable) and falling into a loop.

## Implementation notes
Nothing specific, a minor change to long-standing code.

## Screenshots
N/A

## Notes to reviewer
N/A

### DIFF
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -112,5 +112,6 @@ Rails.configuration.to_prepare do
     DONOTREPLY@3csharedservices.vuelio.co.uk
     D&TCDIO_Office@justice.gov.uk
     FOI.Enquiries@ukaea.uk
+    mail@sf-notifications.com
   )
 end


garethrees (approved):
Didn't even think of this. Nice one!

PR #907, opened 2021-10-29
Donation icon added

lucascumsille:
I added a donation icon(@2 and smaller size).

Connected to https://github.com/mysociety/alaveteli/issues/6547


app/assets/stylesheets/responsive/custom.scss
```diff
@@ -1917,4 +1917,10 @@ $color_asktheeu_dark_green: #C4DDB9;
     display: none;
     visibility: hidden;
   }
-}
\ No newline at end of file
+}
+
+.act-link-icon {background-size: 144px 16px;}
```
zarino, line 1922: We usually try to keep CSS declarations on separate lines, even when there’s only one in the ruleset. That’d make this ruleset look like:

```
.act-link-icon {
  background-size: 144px 16px;
}
```

app/assets/stylesheets/responsive/custom.scss
```diff
@@ -1917,4 +1917,10 @@ $color_asktheeu_dark_green: #C4DDB9;
...
+
+.act-link-icon--donation {
+ background-position: -128px 0;
```
zarino, line 1925: I think the rest of this file uses a 2 space indent, rather than the 1 space at the start of this line.

app/assets/stylesheets/responsive/custom.scss
```diff
@@ -1917,4 +1917,10 @@ $color_asktheeu_dark_green: #C4DDB9;
...
+}
```
zarino, line 1926: Missing a newline at the end of this file.

lib/views/request/_next_actions.html.erb
```diff
@@ -7,3 +7,9 @@
                 AnalyticsEvent::Action::WRITE_TO_THEM_EXIT,
                 :label => 'Request sidebar') %>
 </div>
+
+<div class="act_link">
+  <i class="act-link-icon act-link-icon--donation"></i>
+  <%= link_to 'Donate to WhatDoTheyKnow',
+        '#' %>
+</div>
```
zarino, line 15: Missing a newline at the end of this file.

zarino (commented):
Hope you don‘t mind @lucascumsille, I came across this PR on the Trello card, and figured I’d leave a few comments, especially since there was nobody else assigned to review this.

Aside from the inline comments, a few general questions:

Are the `donation.png` / `donation@2.png` images used anywhere? If not, they probably shouldn’t be included.

Not sure whether you’ve already discussed this with @gbp / @garethrees but I wonder whether the sprite change (new images, and updated CSS) should be made in core, so other installs could benefit from the new icon too? Would mean some un-used data in core (since the core files won’t include a donation `act_link`) but maybe other installs might eventually like to use the icon themselves!

garethrees (commented):
Thanks @zarino!

Yes, I think we'll want these in Alaveteli core as suggested.

I was going to then add something like…

```erb
<% if AlaveteliConfiguration.donate_url %>
  <%= link_to _("Donate to {{site_name}}", site_name: site_name), AlaveteliConfiguration.donate_url class: 'icon donate-icon' %>
<% end %>
```

…to the next actions sidebar partial so that any Alaveteli with the donation URL configured would benefit from this. There's a bit of subtlety around this (due to it being erb, localisable, etc). I didn't expect this as your first Pull Request, so don't worry! But I'm sure @gbp or I can take you through it pretty easily, even if you only observe.

zarino:
Ah! Putting it _all_ into core alaveteli, hidden behind a config variable, makes even more sense.

lucascumsille:
Thank you @zarino @garethrees I appreciate it.
I'll have that in mind on my next pull request

garethrees:
@lucascumsille could you make this pull request over at https://github.com/mysociety/alaveteli? Don't worry about the HTML part, but the CSS would be helpful. Cheers!

lucascumsille:
> @lucascumsille could you make this pull request over at https://github.com/mysociety/alaveteli? Don't worry about the HTML part, but the CSS would be helpful. Cheers!

No problem Gareth. I just did the pull request on Alaveteli.

Issue #931, opened 2021-11-14
Prevent alerts on the internal_admin_user account being turned off


RichardTaylor:
The WhatDoTheyKnow admin team use alerts set up on the internal_admin_user to alert the team to bounce / error messages. 

These keep getting turned off. 

We don't know how. Perhaps someone has a system which automatically follows links in emails (sounds a bit risky!). 

This is a ticket, which I don't really expect to be fixed, suggesting, making them impossible to turn off.

garethrees:
Could patch [`User::EmailAlerts.disable_by_token`](https://github.com/mysociety/alaveteli/blob/0416f6f3e2932c14fba7f88d1b79e7d445ba2f36/app/models/user/email_alerts.rb#L9-L16) in whatdotheyknow-theme like:

```diff
diff --git a/app/models/user/email_alerts.rb b/app/models/user/email_alerts.rb
index 591c4a512..6d6e6474c 100644
--- a/app/models/user/email_alerts.rb
+++ b/app/models/user/email_alerts.rb
@@ -43,6 +43,7 @@ def initialize(user)
     #
     # Returns a Boolean.
     def disable
+      raise 'nope' if user.url_name == 'internal_admin_user'
       user.update(receive_email_alerts: false)
     end
```

PR #932, opened 2021-11-15
[#931] Prevent disabling of Internal Admin email alerts

garethrees:
For some reason we keep seeing email alerts getting disabled for the
Internal Admin User on WhatDoTheyKnow.

We think this is somehow related to the Unsubscribe link rendered via
`app/views/track_mailer/event_digest`. Perhaps someone is mistakenly
clicking it, or an email client is pre-emptively clicking it to cache
content? Who knows!?

This commit checks that the user is _not_ the Internal Admin before
disabling. I used a new module with `prepend` so that we can call
`super` to rely on core behaviour, rather than having to keep the main
method definition in sync.

Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/931.

To test:

```ruby
admin_token = User::EmailAlerts.token(User.internal_admin_user)
app.users_disable_email_alerts_url(token: admin_token)
# => click link - exception raised

user_token = User::EmailAlerts.token(User.first)
app.users_disable_email_alerts_url(token: user_token)
# => click link - alerts disabled
```

### DIFF
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -115,4 +115,18 @@ Rails.configuration.to_prepare do
     mail@sf-notifications.com
     Paul.D.O'Shea@met.police.uk
   )
+
+  User::EmailAlerts.instance_eval do
+    module DisableWithProtection
+      def disable
+        if user.url_name == 'internal_admin_user'
+          raise "Email alerts should not be disabled for #{user.name}!"
+        end
+
+        super
+      end
+    end
+
+    prepend DisableWithProtection
+  end
 end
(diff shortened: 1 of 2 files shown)


gbp (approved):
This seems to works. Not sure about raising an exception as that might mean yet more  emails to go through but volume should be low and might help identify how this is happening.

garethrees:
> This seems to works. Not sure about raising an exception as that might mean yet more emails to go through but volume should be low and might help identify how this is happening.

Yeah, was intentionally meant as a diagnostic tool for the issue. Once we understand it then we can probably do something better (either render a `403` or somehow lock the attribute on that user.

Issue #939, opened 2021-12-02
Add Somerset West and Taunton Deane Council to model_patches.rb


mdeuk:
This Council issues replies from an AchieveForms  / Granicus system that is configured to send out using a no-reply address.

We should add it to model_patches.rb to prevent our users getting stuck in a loop.

no-reply-AT-somersetwestandtaunton-DOT-gov-DOT-uk

Example message is imsg 1901924 (request [799145](https://www.whatdotheyknow.com/request/information_request_re_the_curre_189))

mdeuk:
This was merged by @gbp on 17/12/21 as https://github.com/mysociety/whatdotheyknow-theme/commit/aeb4993650fb06801b87b6120f01dc8c8648bbcc

PR #940, opened 2021-12-02
[#939] Add Somerset West and Taunton Deane to model_patches.rb

mdeuk:
## Relevant issue(s)
#939 

## What does this do?
This change adds the no-reply address used by Somerset West & Taunton Deane council to model_patches.rb

## Why was this needed?
The council issues their replies from a no-reply address, which can get our users stuck in a loop.

## Implementation notes
Nothing specific, it's the usual tried and tested code.

Of slight note perhaps - the address directly above has an apostrophe in it, I'm not certain whether or not this needs to be escaped. I assume not, given the code hasn't malfunctioned, but I'm not 100% certain!

## Screenshots
N/A

## Notes to reviewer
N/A

### DIFF
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -115,6 +115,7 @@ Rails.configuration.to_prepare do
     mail@sf-notifications.com
     Paul.D.O'Shea@met.police.uk
     no-reply@somersetwestandtaunton.gov.uk
+    csfinanceplanning&performance.briefingteam@hmrc.gov.uk
   )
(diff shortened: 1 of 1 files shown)


mdeuk:
I have added to this change, retrospectively, a second address - for HMRC, as noted in the below issue.

It would be good to see a fix for https://github.com/mysociety/alaveteli/issues/3465 in the future, so that we may do away with this workaround.

Fixes #953

Issue #876, opened 2021-08-25
Tweak to help page section on: "Removal of personal information from requests and responses"


RichardTaylor:
> We will consider any such notice, and balance it against any public interest in publishing the material. There is some guidance about this [on the ICO’s website](https://ico.org.uk/for-organisations/guide-to-data-protection/guide-to-the-general-data-protection-regulation-gdpr/individual-rights/right-to-erasure/). 

should state

> We will consider any such notice, and balance it against [our legitimate interests in processing the material](https://www.whatdotheyknow.com/help/privacy#legal_basis). There is some [guidance about this on the ICO’s website](https://ico.org.uk/for-organisations/guide-to-data-protection/guide-to-the-general-data-protection-regulation-gdpr/individual-rights/right-to-erasure/).

RichardTaylor:
Linking to #877 which includes this proposal but is broader.

PR #957, opened 2021-12-21
[#876] Add link to legitimate interests basis

garethrees:
Clarify how we assess removal of information.

Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/876.


garethrees:
@gbp Can you just check this looks okay and merge if so (looks like I made it via web UI without running it)

gbp (approved):

PR #960, opened 2021-12-24
Add Lincs Police no reply address

sallytay:
Some replies are coming from this address but it doesn't accept emails to it. For example https://www.whatdotheyknow.com/request/vehicle_fleet_information_and_go_109?nocache=incoming-1941007#incoming-1941007

### DIFF
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -116,6 +116,7 @@ Rails.configuration.to_prepare do
     Paul.D.O'Shea@met.police.uk
     no-reply@somersetwestandtaunton.gov.uk
     csfinanceplanning&performance.briefingteam@hmrc.gov.uk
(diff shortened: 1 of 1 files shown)


mdeuk:
> Some replies are coming from this address but it doesn't accept emails to it. For example https://www.whatdotheyknow.com/request/vehicle_fleet_information_and_go_109?nocache=incoming-1941007#incoming-1941007

I wonder if this is being caused by an incorrect setting on their email system, rather than an active attempt to refuse emails - we've seen this a couple of times with Office 365.

garethrees:
> I wonder if this is being caused by an incorrect setting on their email system, rather than an active attempt to refuse emails - we've seen this a couple of times with Office 365.

Could well be – might be worth reaching out.

PR #962, opened 2021-12-26
[#961] Change wording on contact form

### LINKED ISSUE #961: Consider changing phrasing on contact form
The contact form on WDTK /help/contact specifically mentions "volunteers" in several places - implying user support mail is being passed directly to the volunteer support team.

This is no longer always strictly correct, as in addition to the core volunteer team, several mySociety staff are also directly involved in the handling of user-support correspondence, and where necessary, in the review of complex matters.

## Proposal ##

I would propose changing the wording slightly, so that we refer to "team" rather than a specific role.

**Why?** this would allow us greater flexibility to not rely on a given role being directly involved with correspondence and would allow us to set expectations more appropriately.

This hopefully isn't a controversial change -the contact form does already use the terms interchangeably; in effect we are simply standardising on one set of phrasing in this scenario.

We actively promote volunteering elsewhere on the site, and have a list of credits, both within the help pages, and in the [mySociety annual report](https://2021.mysociety.org/), therefore I don't think we are losing anything by adopting a more common set of phrasing - we already do this within the support mailbox itself (e.g. as per #884)

mdeuk:
## Relevant issue(s)
Fixes #961 

## What does this do?
This change replaces "volunteers" with "team", along with minor phrasing changes, on /help/contact and in the relevant partial for the contact form.

## Why was this needed?

This is needed to better reflect the increased diversity of the team that handles user support correspondence, and the changing ways in which WhatDoTheyKnow / WhatDoTheyKnow Pro is administered. This is discussed in greater detail on the linked issue.

## Implementation notes
Nothing controversial, this is a series of minor changes to established code.

## Screenshots
N/A

## Notes to reviewer
N/A

### DIFF
--- a/lib/views/help/_contact_form.html.erb
+++ b/lib/views/help/_contact_form.html.erb
@@ -6,7 +6,7 @@
         <%= f.check_box :understand, :required => true %>
         <label for="contact_understand" class="form_label">
             I understand that WhatDoTheyKnow is <strong>not</strong> run by the
-            government, and the WhatDoTheyKnow volunteers <strong>cannot</strong>
+            government, and the WhatDoTheyKnow team <strong>cannot</strong>
             help me with personal matters relating to government services.
         </label>
     </p>
@@ -36,7 +36,7 @@
 
     <p>
         <label class="form_label" for="contact_message">
-            My message to the WhatDoTheyKnow volunteers:
+            My message to the WhatDoTheyKnow team:
         </label>
         <%= f.text_area :message, :rows => 10, :cols => 60, :required => true %>
     </p>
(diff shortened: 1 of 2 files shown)


garethrees (approved):
Agree with the proposal and looks great to me!

RichardTaylor (approved):
This is fine. As noted we do want to promote the work done by volunteers on the service, but that doesn't stop us from getting the wording correct and consistent here.

garethrees:
@sallytay any contrary thoughts before we get this merged?

sallytay:
@garethrees No all looks good to me to

PR #1004, opened 2022-02-22
Add public body questions for Social Security agencies

### LINKED ISSUE #745: Reconsider notes on DWP page
Moved in from the support inbox as a ToDo.

https://www.whatdotheyknow.com/body/dwp

DWP page notes give a link to an INF2 form from 2017 - Consider if to replace that with a link to - https://www.whatdotheyknow.com/request/the_inf2_form (or similar newer version of the form)

> As far as I can see, it's the same form; however perhaps a newer date would be more "attractive" for our users.
> 
> Ideally, we'd want to refresh the information text on that body page - it is rather unwieldy, and I suspect a fair proportion of our users don't read it, meaning we see a number of requests that relate very specifically to an individual's own circumstances. Some of this content could be special category information as it can relate to a person's health and wellbeing.
> 
> We do have a ticket for adding an interstitial to this body (similar to the version used for Home Office requests[1]) - #722. This may help us to stop some of these duplicate requests. Any feedback is welcome on this - e.g. what specific information we might want to include.
> 
> On the point of the DWP - we should consider whether or not we are serving our users best interests by listing the "Child Support Agency" and "Child Maintenance Options" separately. At one point these bodies were executive agencies, similar to JobCentre Plus, however this no longer appears to be the case.
> 
> In some cases we're seeing users inadvertently make requests which are being considered vexatious [2,3,4] as they are sending to CSA, CMO, and DWP separately - without realising they are actually the same entity.
> 
> This creates confusion, and I wonder if we'd be better serving our users if we mark those bodies as defunct and place a note suggesting that their request be made to the DWP instead. We could place the "brand name" of those entities in the information text for the DWP to tie things together.
> 
> [1] https://www.whatdotheyknow.com/new/home_office (view in incognito / private browsing)
> [2] https://www.whatdotheyknow.com/request/managing_information_disclosure_27
> [3] https://www.whatdotheyknow.com/request/managing_information_disclosure_28
> [4] https://www.whatdotheyknow.com/request/managing_information_disclosure_55

mdeuk:
## Relevant issue(s)
Fixes #745 
Partly solves #722 

## What does this do?

This implements a series of public body questions for the three Social Security agencies in the UK:

- Department for Communities (Northern Ireland) `(dfc - 57878)`
- Department for Work and Pensions `(dwp - 24)`
- Social Security Scotland `(social_security_scotland - 89236)`

It also refactors the questions we have for the Home Office `(home_office - body 7)`, and improves the generic answer that we have for said body, by including some specific pointers.

Key themes:

- Claim benefits / state pension
- Ask about a benefit claim
- Ask for a copy of records (e.g. RoAR / SAR)
- Ask about Visas
- Ask about Passports

## Why was this needed?

We consistently see users contacting these bodies in regard to matters which concern their own personal circumstances and in no way relate to a FOI or EIR request. This can result in personally identifiable information, which is often hard to remove, being published.

Being able to signpost users to relevant _external resources_ and dissuade them from misusing our service _may_ help reduce misuse, which takes significant administrative resources to resolve.

## Implementation notes
This will work best if https://github.com/mysociety/alaveteli/pull/6807 is implemented, as the questions tie into the simplified heading that has been used here.

The public body name for Social Security Scotland (e.g. social_security_scotland) has been shortened to `socsecscot` in the code - this has been tested on a dev instance and showed no issues.

There are a couple of generic boilerplate statements (`generic_deny_boilerplate`, `generic_deny_askcouncil`,`generic_deny_gdpr_rightofaccess`) to save repetition and improve maintainability. The first is essentially a facsimile of the original misuse text from the Home Office, which has been refactored and simplified.

## Screenshots

A short video showing some of the options:

https://user-images.githubusercontent.com/249418/155139904-ed18da96-b7ea-43be-9b5e-8263a7f85c3a.mov

Home Office:
- 
- 

DfC:
- 
- 

Social Security Scotland
- 
- 

## Notes to reviewer

There is a bug of sorts - but it appears to exist in core Alaveteli - when a radio button is clicked, we don't automatically focus the browser on the relevant text, meaning the user needs to scroll. This isn't a _massive_ thing, but it might be slightly more intuitive if we did that.

Some care has been taken to ensure our responses are neutral, in that we don't express an opinion regarding a public bodies policies or procedures - the links to external resources, which are from reputable organisations such as Citizens Advice, can do that much better than we can - as they have the resources to let them keep data _up-to-date_.

Some of the off-site links are, unfortunately, rather long - sorry about that.

### DIFF
--- a/lib/public_body_questions.rb
+++ b/lib/public_body_questions.rb
@@ -42,6 +42,7 @@ Rails.configuration.to_prepare do
       </p>
     HTML
   )
+  ## Generic boilerplate templates for reuse
 
   ## LSE
 
@@ -134,7 +135,7 @@ Rails.configuration.to_prepare do
     HTML
     # we reuse this multiple times with #{generic_deny_boilerplate}
   )
-  
+
   generic_deny_askcouncil  = _(<<-HTML.strip_heredoc.squish
       <p>
         Your local Council might also offer a Welfare Rights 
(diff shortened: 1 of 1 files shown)


garethrees:
Thanks for this! It's probably going to take us a little while to review due to annual leave and ongoing LSE issues, but we'll do our best 🙏

mdeuk:
> Thanks for this! It's probably going to take us a little while to review due to annual leave and ongoing LSE issues, but we'll do our best 🙏

It’s not quite ready yet - there’s a further commit coming (hopefully tonight) to add some contest to links.

Sorry it’s quite a hefty script though - there’s a fair bit of nuance which prevents reuse of some elements, but I’ve tried to be as generic as possible.

PR #1009, opened 2022-02-24
Add LSE questions

### LINKED ISSUE #1008: Implement public body questions for 3 organisations
Based on an established pattern  of misuse, we will need to consider implementing some public body questions for bodies 525, 1294, and 12848.

Draft: https://docs.google.com/document/d/1yvre2clIqO3fhzCU2hWv-L3g1Y61yQLeEkkFG4qu2Iw/edit

mdeuk:
**Queries to: https://www.whatdotheyknow.com/help/contact**

## Relevant issue(s)
Fixes #1008 

## What does this do?
This adds public body questions for bodies 525, 1294, and 12848.

## Why was this needed?
To reduce a misuse vector, and improve the quality of our advice to site users.

## Implementation notes
Copy has been done relatively quickly, feel free to suggest  any amendments.

This will work best once https://github.com/mysociety/alaveteli/pull/6807 is deployed, but it's by no means compulsory - it's just a heading difference.

## Screenshots
N/A

## Notes to reviewer

As discussed on review@ - this is a follow-up to the mockup version.

### DIFF
--- a/lib/public_body_questions.rb
+++ b/lib/public_body_questions.rb
@@ -1,7 +1,17 @@
 Rails.configuration.to_prepare do
   next unless ActiveRecord::Base.connection.data_source_exists?(:public_bodies)
 
+  ## Specify public bodies which should be evaluated
+
   home_office = PublicBody.find_by_url_name('home_office')
+  ## LSE
+  lse_tsai = PublicBody.find_by_url_name('lse')
+  uol_tsai = PublicBody.find_by_url_name('university_of_london')
+  uolww_tsai = PublicBody.find_by_url_name('university_of_london_worldwide')
+
+  ## Generic boilerplate templates for reuse
+
+  ## public body specific templates
 
   home_office_deny_response = _(
     <<-HTML.strip_heredoc.squish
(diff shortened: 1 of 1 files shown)


mdeuk:
I would add - I've tested this on a sandbox and it _works_ - but it does feel slightly 'icky'. 

I'd hoped to have this apply using tags, which would have allowed a better solution, but I couldn't get that to work.

I blame a lack of coffee. If there's a more elegant way to do this please let me know 🙏

gbp:
image

image

image

gbp (commented):
I have tested and this looks to work correctly - see screenshots. Well done @mdeuk for making sense of this Ruby code. I would be happy to merge this although is this still an ongoing problem. @garethrees ?

garethrees (approved):
Yeah I think we should merge.

While the misuse has subsided due to the technical measures we've put in place, we'd like to ease those as they are catching a handful of genuine users.

I think this makes our house rules really clear so that if we see further misuse we'll know it's not through misunderstanding.

PR #1020, opened 2022-03-01
Add no reply address for Cambridgeshire County Council

sallytay:
When information is sent via the file transfer service it comes from a no reply email address. This address then gets picked up as the reply address for WhatDoTheyKnow users creating a bounce message rather than going through to the authority


mdeuk (approved):
Verified against imsg 1978067 - looks good to me 👍

Issue #1035, opened 2022-03-04
Confusing phrasing of Legal basis for processing


garethrees:
Our section on [Legal basis for processing](https://www.whatdotheyknow.com/help/privacy#legal_basis) states:

> In most cases our legal basis for processing personal information is "legitimate interest"…

It then goes on to say (em mine)…

> We believe that our processing of our users’ data is as they would expect when they use our service. If you use WhatDoTheyKnow to make a FOI request, you are **consenting** to your data being processed as described on this page.

I don't think this is accurate, since we process all GDPR RTE requests under the legal basis of Legitimate Interests.

RichardTaylor:
I agree. 

We could just delete:

>, you are consenting to your data being processed as described on this page.

PR #1037, opened 2022-03-04
Fix phrasing of legal basis for processing

garethrees:
We process all GDPR RTE requests under the legal basis of Legitimate Interests,
so using the word "consent" in this particular context is a little confusing given
it's a specific term for another legal basis for processing.

Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/1035


mdeuk (approved):
LGTM 👍

PR #1063, opened 2022-03-16
[#1059] Add outlook recall information to /help/officers

### LINKED ISSUE #1059: Help page section on Outlook "Recall" messages
This comes up often enough that we ought to have some pre-existing wording around handling these.

mdeuk:
## Relevant issue(s)
Fixes #1059 

## What does this do?

This adds text to /help/officers (in English only) detailing why "recall" messages sent through Microsoft Outlook (Microsoft Exchange email systems) will fail to work.

## Why was this needed?

Public bodies try to use the "recall" function to remove information from WDTK; however, this is not supported on the majority of platforms - ours included.

## Implementation notes
Straight HTML - there's an annoyingly _massive_ Microsoft link which is breaking line length - Microsoft haven't made a short url available, we could create one if really needed.

## Screenshots
N/A

## Notes to reviewer

This is probably the first step - the ticket upstream (https://github.com/mysociety/alaveteli/issues/6884) will allow us to consider something more substantive in the Alaveteli codebase to handle these nuisance messages in a more elegant way.

### DIFF
--- a/lib/views/help/officers.html.erb
+++ b/lib/views/help/officers.html.erb
@@ -378,18 +378,19 @@
         both the sender and recipient are on the same email system.
       </p>
       <p>
-      <p>
         This functionality is unlikely to work on other platforms; and it is not a 
         function that is supported on WhatDoTheyKnow. Further information about 
         this function, and its limitations, may be found on the 
         <a href="https://support.microsoft.com/en-us/office/recall-or-replace-an-email-message-that-you-sent-35027f88-d655-4554-b4f8-6c0729a723a0?ui=en-us&rs=en-us&ad=us"
         title="Find out about message recalls on the Microsoft website">
         Microsoft website</a>.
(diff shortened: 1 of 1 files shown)


mdeuk:
> At [#1059 (comment)](https://github.com/mysociety/whatdotheyknow-theme/issues/1059#issuecomment-1069540112) William has proposed switching "and we'll help" to "and we'll look into it".
> 
> It does appear sensible not to raise expectations that we'll actually help (or do what the person making the report wants).

"help" was intended to be rather open ended by default - but this is a reasonable change to make. Duly updated in https://github.com/mysociety/whatdotheyknow-theme/pull/1063/commits/6a1b9b7e49578e725dd4c37701ae6d3d939917ad.

garethrees:
Awesome, thanks. Squashed and merged as 0dda0e3.

PR #1096, opened 2022-03-24
Add patch to ActiveStorage::Blob#delete

gbp:
Prevent deletion of variants. We don't currently use variants and this
causes timeouts (when doing a remote globs to find variant files to be
deleted) when using a ActiveStorage SFTP service.

See: https://github.com/rails/rails/blob/6-1-stable/activestorage/app/models/active_storage/blob.rb#L287-L290

### DIFF
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -134,4 +134,14 @@ Rails.configuration.to_prepare do
 
     prepend DisableWithProtection
   end
+
+  ActiveStorage::Blob.class_eval do
+    def delete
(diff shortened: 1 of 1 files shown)


gbp:
@garethrees happy for this to be patched here instead of in core? The timeouts seems to be an issue with how much data WDTK has rather than an issue which might effect other installs.

garethrees (approved):
> The timeouts seems to be an issue with how much data WDTK has rather than an issue which might effect other installs.

👍  makes sense.

Issue #1138, opened 2022-04-14
Accessibility issues


lucascumsille:
Hi @garethrees last sprint we talked about me, checking the WhatDoTheyKnow theme. These are some of the issues I have found. Let me know if you want me apply some fixes to them.

#### Blue Button
<img width="330" alt="Screenshot 2022-04-14 at 08 58 50" src="https://user-images.githubusercontent.com/13790153/163359830-5be08c82-de54-4423-b252-1cda465509f4.png">

#### Navigation bar
<img width="855" alt="Screenshot 2022-04-14 at 09 00 55" src="https://user-images.githubusercontent.com/13790153/163359963-6142da69-8f94-4f3f-aba9-56a2bab837c1.png">

### Homepage
<img width="916" alt="Screenshot 2022-04-14 at 08 57 12" src="https://user-images.githubusercontent.com/13790153/163360066-552d0341-ddf4-4d55-81ac-168bd7316356.png">

### Dashboard (en/alaveteli_pro/)
<img width="855" alt="Screenshot 2022-04-14 at 09 02 25" src="https://user-images.githubusercontent.com/13790153/163360534-06de4e66-b0fb-4823-ab29-6cee48a7f061.png">

### en/alaveteli_pro/info_requests/new
I would also give the warning a yellow or red colour, so it pops out more. Plus some vertical alignment so the the lock and text are at the same height.

<img width="497" alt="Screenshot 2022-04-14 at 10 52 17" src="https://user-images.githubusercontent.com/13790153/163360894-473ceabc-0d1e-433a-93e9-97b611dd4095.png">
<img width="497" alt="Screenshot 2022-04-14 at 10 53 04" src="https://user-images.githubusercontent.com/13790153/163361038-4c6f741c-6eb0-4039-a45b-683067277567.png">

### en/request/the_cost_of_boring_two#incoming-3
Not a major thing

"Help" button 
Possible solution
<img width="1029" alt="Screenshot 2022-04-14 at 11 11 36" src="https://user-images.githubusercontent.com/13790153/163364533-7a127f35-062b-4b54-b1b9-3b040397fab9.png">
`
    font-size: 10px;
    text-transform: uppercase;
    font-weight: 700;
`

Current state
<img width="1029" alt="Screenshot 2022-04-14 at 11 12 00" src="https://user-images.githubusercontent.com/13790153/163364515-a2cf36c4-f29a-42af-b711-488ac1b5dcb0.png">

[...]

PR #1146, opened 2022-04-21
Accessibility issues: WDTK

lucascumsille:
This pull request contains some fixes for accessibility issues found in the WDTK theme.

### $color_blue is now a darker, richer version that passes the contrast test:

Screenshot 2022-04-21 at 07 36 10
Screenshot 2022-04-21 at 07 35 45
Screenshot 2022-04-21 at 07 35 28

### Items on subnav pass contrast test
Screenshot 2022-04-21 at 08 27 40

### Secondary button a bit darker

Screenshot 2022-04-21 at 08 50 38

### Increase vertical padding for list items on help page

Screenshot 2022-04-21 at 07 37 53

### When creating a new request the footer notes now pass the contrast test
Screenshot 2022-04-21 at 07 40 26

Fixes part of: #1138 
The rest of the issues, will be fixed on the Alaveteli core.

### DIFF
--- a/app/assets/stylesheets/responsive/custom.scss
+++ b/app/assets/stylesheets/responsive/custom.scss
@@ -1,5 +1,5 @@
 $color_orange: #f4a140;
-$color_blue: #419BE8;
+$color_blue: #2078C2;
 $color_green: #62b356;
 $color_yellow: #ffd836;
 $color_red: #e04b4b;
@@ -14,7 +14,7 @@ $color_white: #ffffff;
 $color_pro_blue: #5A7687;
 
 $mysociety-banner: $color_black;
-$link-color: darken($color_blue, 20%);
+$link-color: darken($color_blue, 10%);
 $body-font-color: $color_black;
 $body-bg: $color_off_white;
 $button-bg: $color_blue;
--- a/app/assets/stylesheets/responsive/custom.scss
+++ b/app/assets/stylesheets/responsive/custom.scss
@@ -311,7 +311,7 @@ a.link_button_green_large {
 }
 
 .form_item_note {
-  color: $color_mid_grey;
+  color: $color_dark_grey;
 }
 
 /* Popups */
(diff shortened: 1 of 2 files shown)


mdeuk:
> This pull request contains some fixes for accessibility issues found in the WDTK theme.
> ### $color_blue is now a darker, richer version that passes the contrast test:
> 
> Screenshot 2022-04-21 at 07 35 45
> 
> ### Items on subnav pass contrast test
> 
> Screenshot 2022-04-21 at 08 27 40

Just to note, and definitely not a review - I actually really like this - it has the secondary advantage of refreshing the design in a subtle way, whilst delivering known access benefits. Lovely stuff 👍

garethrees (commented):
🤌 I like it!

While the previous version wasn't an _exact_ match, it was much closer to the [mySociety brand colours](https://wiki.mysociety.org/wiki/Logo_and_Branding). @zarino do you have an opinion on this?

app/assets/stylesheets/responsive/custom.scss
```diff
@@ -1399,6 +1399,7 @@ p.locale-list-trigger {
   ul {
     list-style: none outside none;
     padding: 0;
+    li {padding: 5px 0;}
```
zarino, line 1402: Minor thing, but I’d format this onto a new line, like all of the other rules nearby:

```scss
  ul {
    list-style: none outside none;
    padding: 0;
    li {
      padding: 5px 0;
    }
  }
```

zarino (approved):
This looks good to me! I agree the darker colour is a bit of a shock at first, but there’s no need to keep it the same as the mySociety blue, and this darker blue makes the contrast of links and the header much better. Deffo worth it. :+1:

gbp:
I'm thinking we merge and deploy this when we start talking about our 20th anniversary. Am I correct in thinking there will be a banner added to WDTK for this?

garethrees:
> I'm thinking we merge and deploy this when we start talking about our 20th anniversary

Sounds good.

> Am I correct in thinking there will be a banner added to WDTK for this?

Looks like its on its way https://github.com/mysociety/Design-Internal-tasks/issues/8#issuecomment-1410057767

Issue #1133, opened 2022-04-09
Welsh version of help page still mentions UKCOD


FOIMonkey:
https://www.whatdotheyknow.com/cy/help/about has "Mae mySociety yn brosiect yr elusen gofrestredig Democratiaeth Dinasyddion Ar-lein y DG." This is no longer correct following the name change and needs updating. 

It could be changed to the following to more closely mirror the English: 

Mae mySociety yn elusen gofrestredig (1076346) ac yn gwmni wedi'i gofrestru yng Nghymru a Lloegr (03277032).

WilliamWDTK:
This is similar to #650, in that it's about Welsh translation.

Also, just to add something you (@FOIMonkey) mentioned on Slack (that I accidentally just posted a (since deleted) message with the search term, rather than using the search box 🤦):

> TIL that the Welsh Government offers individuals/charities/businesses 500 words of Welsh translation for free per calendar month with a 4 working day turnaround. Might be useful: https://businesswales.gov.wales/heloblod/helo-blod

[(source)](https://mysociety.slack.com/archives/CNH83HJP4/p1647848574165539)

I've also found [Charity Translators](https://www.charitytranslators.org/), it's not clear how much / exactly what they offer (it might just be guidance, rather than actually offering translation), but we could look into them further.  There is [PerMondo](http://www.permondo.eu/non-profit-organisations/), as well, but the main sponsor company doesn't offer Welsh translation (though that's not to say they don't have another volunteer that can help).

FOIMonkey:
Great suggestions that I think are definitely worth looking into. The above was my best effort, but I've only been learning Welsh for a few months.

mdeuk:
> Great suggestions that I think are definitely worth looking into. The above was my best effort, but I've only been learning Welsh for a few months.

I've set up a Helo Blod account and added the details to the Wiki. Thanks @FOIMonkey and @WilliamWDTK for this - improving our coverage here is very helpful indeed.

PR #1199, opened 2022-05-16
[#1133] Fix company names on /cy/help/about

garethrees:
Still mentions UKCOD, in Welsh.

Thanks to @FOIMonkey for the translation.

Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/1133.

Screenshot 2022-05-16 at 09 28 10


mdeuk (changes requested):
LGTM - save for one niggle, can we fix the URL here?

Issue #1267, opened 2022-07-08
Link to the Charity Commission for Northern Ireland database


mdeuk:
From a recent discussion on Slack, it was noted by @RichardTaylor that:

> I had forgotten we had magic to detect the type of Charity number and wondered if I needed to so something special with tagging Scottish charities - and found I don't.

We do this by handling the 'charity' tag through special functionality in [/lib/views/public_body/_more_info.html.erb](https://github.com/mysociety/whatdotheyknow-theme/blob/master/lib/views/public_body/_more_info.html.erb) - however, this only works for charities that appear on the registers of the Charity Commission for England and Wales (CCEW), or the Office of the Scottish Charity Regulator (OSCR). We should add the Charity Commission for Northern Ireland (CCNI) to this.

The format for CCNI is broadly similar to CCEW (it appears to be the same software) - with a slight nuance. Charities in NI use an alphanumeric identifier (e.g. [NIC102724](https://simoncommunity.org/legal)), but the database only accepts the numeric identifier - so we will need to accommodate this.

https://www.charitycommissionni.org.uk/charity-search/?searchText=#{ tag_value } would seem to be a valid structure to use.

garethrees:
If I understand correctly, all we'd need is a minor change to the sidebar and to tag NI charities as `NIC:102724`?

```diff
diff --git a/lib/views/public_body/_more_info.html.erb b/lib/views/public_body/_more_info.html.erb
index 17a111c..8ee6c67 100644
--- a/lib/views/public_body/_more_info.html.erb
+++ b/lib/views/public_body/_more_info.html.erb
@@ -17,6 +17,9 @@
     <% if tag_value.match(/^SC/) %>
       <%= link_to _('Charity registration'),
                     "https://www.oscr.org.uk/about-charities/search-the-register/charity-details?number=#{ tag_value }" %><br>
+    <% elsif tag_value.match(/^NIC/) %>
+      <%= link_to _('Charity registration'),
+                    "https://www.charitycommissionni.org.uk/charity-search/?searchText=#{ tag_value }" %><br>
     <% else %>
       <%= link_to _('Charity registration'),
                     "https://register-of-charities.charitycommission.gov.uk/charity-search/-/results/page/1/delta/20/keywords/#{ tag_value }" %><br>
```

garethrees:
[...]

PR #1284, opened 2022-07-14
[#1267] Add special link for NI charities

garethrees:
Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/1267.

### DIFF
--- a/lib/views/public_body/_more_info.html.erb
+++ b/lib/views/public_body/_more_info.html.erb
@@ -17,6 +17,9 @@
     <% if tag_value.match(/^SC/) %>
       <%= link_to _('Charity registration'),
                     "https://www.oscr.org.uk/about-charities/search-the-register/charity-details?number=#{ tag_value }" %><br>
+    <% elsif tag_value.match(/^NIC/) %>
+      <%= link_to _('Charity registration'),
+                    "https://www.charitycommissionni.org.uk/charity-search/?regId=#{ tag_value.gsub(/^NIC/, '') }" %><br>
     <% else %>
       <%= link_to _('Charity registration'),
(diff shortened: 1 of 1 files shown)


lib/views/public_body/_more_info.html.erb
```diff
@@ -17,6 +17,9 @@
     <% if tag_value.match(/^SC/) %>
       <%= link_to _('Charity registration'),
                     "https://www.oscr.org.uk/about-charities/search-the-register/charity-details?number=#{ tag_value }" %><br>
+    <% elsif tag_value.match(/^NIC/) %>
+      <%= link_to _('Charity registration'),
+                    "https://www.charitycommissionni.org.uk/charity-search/?searchText=#{ tag_value.gsub(/^NIC/, '') }" %><br>
```
garethrees, line 22: Was just about to merge this but then noticed we _could_, it seems, link direct to the Charity Details page.

e.g. for `charity:NIC102724`, we could link to https://www.charitycommissionni.org.uk/charity-details/?regId=102724 instead of https://www.charitycommissionni.org.uk/charity-search/?searchText=102724.

Any reason not to do this @mdeuk?

mdeuk:
```suggestion
                    "https://www.charitycommissionni.org.uk/charity-search/?regId=#{ tag_value.gsub(/^NIC/, '') }" %><br>
```

**I can't see any reason not to do this!**

I've ran through the database for NI and can't see any reason why this _**wouldn't**_ work. We didn't do it for CCEW as the identifiers in their database are awkward to translate (e.g. mySociety's reg of  1076346 becomes [3078648](https://register-of-charities.charitycommission.gov.uk/charity-search/-/charity-details/3078648)).

Good catch 👍

mdeuk (approved):

PR #1289, opened 2022-07-14
[#1276 / #1286] Improve stylng of forms in /help/contact and improve copy

### LINKED ISSUE #1276: Review use of bullet points / styling in contact form
When the various contact forms on /help/contact are rendered, we are rendering the headline text in bullet points, and then rendering the form itself in another bullet point.

### Why is it an issue?

Rendering a bullet point next to the form looks a little peculiar, and slightly less polished than our UX usually is. This is probably nitpicking, and it might purely be an aesthetic complaint, rather than broken functionality.

Image of the WhatDoTheyKnow volunteer contact form, showing that there are three visible bullet points - two for the listing of options, and one for the form itself. An orange circle has been drawn over the bullet point for the form, and the annotation 'Why?' has been added

### Possible solution

We already have a css selector which will [remove visible bullet points](https://github.com/mysociety/alaveteli/blob/master/app/assets/stylesheets/responsive/_help_style.scss) from unstructured lists, but applying this this also removes the padding on the left margin, which makes the form slightly less logical in appearance.

Image of the WhatDoTheyKnow volunteer contact form, showing that the bullet points have been removed from visible styling, and there is no no padding on the left hand side. An orange bracket has been on the left hand side, indicating where there should be space, and the annotation 'Lack of space to breathe' has been overlaid

### Suggested fix

This is a relatively obscure usage, so if adding to the core stylesheet is deemed to be overkill, a hack would be to apply styling at a local level (e.g. `<li style="list-style-type:none;">`). This would, perhaps, only need to be applied to the list item containing the form itself - so as not to remove the visible bullet from areas where we might wish to retain it.

Example:
Image of the WhatDoTheyKnow volunteer contact form, showing that the bullet points for the list items still exist, but that the bullet point appearing at the left hand side of the form has been removed. A green circle has been drawn where the bullet point ought to have been, and the annotation 'Bullet point removed, space preserved' has been overlaid.

This isn't necessarily a perfect solution, so I'll open the floor to the design gurus who might have a better idea than me.

### LINKED ISSUE #1286: Differentiate pro/non-pro for all parts of /help/contact
Goal 3, aka "I need help writing a freedom of information request" states "Directly contact the team that run WhatDoTheyKnow" - which is slightly inconsistent with the copy in Goal 5, which differentiates between pro and non-pro uses.

Adding a link to /help/requesting might also be of benefit, as there's a good chance there is already an answer on there.

mdeuk:
## Relevant issue(s)
Fixes #1276
Fixes #1286

## What does this do?
This adds a slight bit of styling to the `<li>` element within the houdini goals where a contact form is rendered.

It also makes an attempt at improving the copy in goal 3 - "I need help writing a freedom of information request" and adds some variation within aforementioned for Pro.

## Why was this needed?

The volunteer contact form on /help/contact introduced a slight oddity of styling, where the form itself was contained in a separate list element of its own, which led to a slightly unsightly bullet point being rendered. This is a kludgy attempt at remedying this, without the expense of adding to the core stylesheet.

We were also somewhat light with the copy in goal 3, which is odd, as we've some pretty comprehensive information in the help pages. It seems wise to link there, and to the blog, as a first port of call.

## Implementation notes
Nothing of real interest here, it's simple html for the most part.

#### original:
Image of the WhatDoTheyKnow volunteer contact form, showing that there are three visible bullet points - two for the listing of options, and one for the form itself. An orange circle has been drawn over the bullet point for the form, and the annotation 'Why?' has been added

#### revised:
Image of the WhatDoTheyKnow volunteer contact form, showing that the bullet points for the list items still exist, but that the bullet point appearing at the left hand side of the form has been removed. A green circle has been drawn where the bullet point ought to have been, and the annotation 'Bullet point removed, space preserved' has been overlaid.

### Writing help
Image of the WhatDoTheyKnow writing help contact form, showing new text which has been added giving suggestions on how the user might find help in writing their request. The full text is within the commit

## Notes to reviewer
It feels like we could cut the if/else for Pro in goal 5 (and by definition, my copy/paste in goal 3) down a bit. It isn't massively problematic though, it's not like maintainability is an issue.

### DIFF
--- a/lib/views/help/contact.html.erb
+++ b/lib/views/help/contact.html.erb
@@ -79,7 +79,8 @@
         <ul>
             <li>
                 Directly contact the team that run WhatDoTheyKnow:
-
+            </li>
+            <li style="list-style-type:none;">
                 <%= foi_error_messages_for :contact %>
 
                 <%= render :partial => "help/contact_form",
@@ -108,7 +109,7 @@
                 involves, see
                 <%= link_to _("our get involved page"), help_volunteers_path %>.
             </li>
-            <li>
+            <li style="list-style-type:none;">
                 <%= foi_error_messages_for :contact %>
 
                 <%= render :partial => "help/contact_volunteer_form",
@@ -143,7 +144,8 @@
                     If your issue isn’t covered by our help pages, you can
                     directly contact the team that run WhatDoTheyKnow:
                 <% end %>
-
+            </li>
+            <li style="list-style-type:none;">
                 <%= foi_error_messages_for :contact %>
 
                 <%= render :partial => "help/contact_form",
--- a/lib/views/help/contact.html.erb
+++ b/lib/views/help/contact.html.erb
@@ -93,8 +93,9 @@
                     run WhatDoTheyKnow Pro. 
                 <% else %>
                    If you still need help, you can contact the team that 
-                    run WhatDoTheyKnow. 
+                    run WhatDoTheyKnow.
                 <% end %>
+                <br>
                 It'd be helpful if you could let us know what information you 
                 are looking for, and which body you think might hold it.
             </li>
--- a/lib/views/help/contact.html.erb
+++ b/lib/views/help/contact.html.erb
@@ -88,13 +88,8 @@
                 <%= link_to _("our blog"), blog_path %>.
             </li>
             <li>
-                <% if @user && @user.is_pro? %>
-                    If you still need help, you can contact the team that 
-                    run WhatDoTheyKnow Pro. 
-                <% else %>
-                   If you still need help, you can contact the team that 
-                    run WhatDoTheyKnow.
-                <% end %>
+                If you still need help, you can contact the team that 
+                run WhatDoTheyKnow.
                 <br>
                 It'd be helpful if you could let us know what information you 
                 are looking for, and which body you think might hold it.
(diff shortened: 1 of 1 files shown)


lib/views/help/contact.html.erb
```diff
@@ -76,9 +76,27 @@
...
+                Our <a href="<%= help_requesting_path %>">help pages</a> have 
+                guidance on how to make requests, including some top tips on 
+                how you can get the best response.
+            </li>
+            <li>
+                You can also find inspiration for your request on 
+                <a href="<%= link_to _("Read blog"), blog_path %>">our blog</a>.
+            </li>
+            <li>
+                <% if @user && @user.is_pro? %>
```
garethrees, line 91: -1 to this conditional; it's the same team, just different people _tend_ to respond based on the current user status. I don't want to have to differentiate this everywhere in the app. The contact form routes correspondence to the right inbox.

mdeuk: I’d rather we didn’t differentiate at all, but we do elsewhere on the page.

I suppose you could take the view that a pro user might be less likely to use this form though… will post a fix up later.

garethrees: > I’d rather we didn’t differentiate at all, but we do elsewhere on the page.

Ah I didn't notice the existing conditional. Agree – would be happy for both to be removed!

mdeuk: > Ah I didn't notice the existing conditional. Agree – would be happy for both to be removed!

Even better, less moving parts is always a maintainability bonus. 😃 

Fixup duly pushed in 479fabcb4eedb3d50f95ee6852bd0ab1fad3c198

gbp:
Merged 6f78fe412be4108b9d684085bc0c598e2c64d92b

PR #1327, opened 2022-08-02
[#1326] Update help pages

### LINKED ISSUE #1326: Links to OSIC and ICO guidance in `/help/unhappy` are out of date
Some of the links within `/help/unhappy`, to guidance on the Information Commissioner, and Scottish Information Commissioner (OSIC) websites are broken. This is because OSIC have a new(ish) website, and the ICO are gradually replacing their PDF guidance.

mdeuk:
## Relevant issue(s)
Fixes #1326 
_#650_

## What does this do?
This is a chunky PR, but in essence it:

- Removes the contact form from `lib/views/help/contact.cy.html.erb` and sets up a partial
- Adds partial `lib/views/help/_history.cy.html.erb`, so that we no longer present the English version on /cy
- Replaces reference to DPA 1998 and briefly covers RtE / RoA (SAR)
- Roughly translates aforementioned
- Fixes a couple of broken links

## Why was this needed?
We had a couple of broken links. At the same time, our Welsh help pages are quite a bit out of sync, so this makes a first attempt at trying to rectify this.

## Implementation notes
Nothing of particular note; however, I would note that although `lib/views/help/contact.cy.html.erb` references the new Volunteer contact form, the partial hasn't been translated, as I believe it would need the specs to be updated. We could remove Houdini goal 4 if necessary.

## Screenshots
n/a

## Notes to reviewer
Apologies - the translations here are machine generated, but wherever possible they have been verified against two different translation engines, and terminology cross-referenced against [BydTermCymru](https://gov.wales/bydtermcymru).

This intentionally only covers the basic parts - we owe it to our Welsh speaking users to ensure that the more complicated parts have had their translations verified (e.g. Helo Blod, or a suitable volunteer).

### DIFF
--- a/lib/views/help/_contact_form.cy.html.erb
+++ b/lib/views/help/_contact_form.cy.html.erb
@@ -1,0 +1,83 @@
+<%= form_for :contact,
+             :url => help_contact_path + '#' +  form_id,
+             :html => {:class => 'contact-form'} do |f| %>
+
+    <p class="contact-form__understand">
+        <%= f.check_box :understand, :required => true %>
+        <label for="contact_understand" class="form_label">
+            Rwy'n deall <strong>nad yw WhatDoTheyKnow yn cael ei redeg gan y 
+            llywodraeth, ac ni fydd tîm WhatDoTheyKnow <b>yn gallu helpu</b> 
+            gyda materion personol sy'n gysylltiedig â gwasanaethau'r llywodraeth.
+        </label>
+    </p>
+    <% if not @user %>
+        <p>
+            <label class="form_label" for="contact_name">Eich enw:</label>    
+            <%= f.text_field :name, :size => 20, :required => true %>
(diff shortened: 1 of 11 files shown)


mdeuk:
I should add, there is also a SocietyWorks element here - the copy for the Pro upsell was outdated. This is covered off in b859a658bac46a790cb7fd3fa22323c7b466ce3e. We still need to translate the /pro landing pages.

gbp (approved):
This was hard to review as there is so much going on. It does seems okay with no issues rendering the pages. The EN changes are all sensible and I'm happy with the approach taken for the CY changes. Thanks

mdeuk:
> This was hard to review as there is so much going on. It does seems okay with no issues rendering the pages. The EN changes are all sensible and I'm happy with the approach taken for the CY changes. Thanks

Sorry about that, I agree there's too much going on in this one - will try not to make that mistake again. 😳

Issue #818, opened 2021-05-25
If a body is tagged with a MapIt ID then show a map of the area covered by the body


RichardTaylor:
eg. if 

https://www.whatdotheyknow.com/body/herefordshire_and_worcestershire_ccg

was tagged

mapit:163965

show a map of the area covered by the CCG on the CCG page. 

https://mapit.mysociety.org/area/163965.html

<img width="1239" alt="map-on-page" src="https://user-images.githubusercontent.com/547695/119505220-a5bdf100-bd64-11eb-9f18-f3c592ffda27.png">

Related: https://github.com/mysociety/alaveteli/issues/235 and https://github.com/mysociety/whatdotheyknow-theme/issues/579

mdeuk:
**I love this idea** - the example you've given is one that'd also work extremely well for Councils, particularly as we look to improve our coverage of Town and Parish councils, where knowing the particular locus can be really, really helpful!

mdeuk:
Just to add, I can see this also being very helpful for our regional pages - the enhanced metadata could allow for a more enriched experience to be generated there with less overall work, and it could also allow for better targetting of pro user requests (as an example: you might wish to make a FOI request to schools in a given local authority area, or to NHS trusts [bearing in mind there is often more than one]).

#416 #715 #716 #717 #747

garethrees:
Just linking to an [email thread](https://groups.google.com/a/mysociety.org/g/alaveteli-dev-private/c/nBbJxj705wo/m/IDItaxXvAQAJ) about the Mapit aspect of this.

There's no embed option in Mapit, but would be relatively easy to render by the sounds of things:

```js
// The minimal HTML you'd need to replicate the map on WDTK would be
// something like: add the Leaflet CSS to the head of the page, and the
// Leaflet JS to the bottom, stick a <div id="map"></div> in the HTML
// with some styling to give it a width/height, then have a bit of extra
// JavaScript that does the same as MapIt to plot the area:

// Set up map with OSM background mapping
var map = new L.Map("leaflet");
map.attributionControl.setPrefix('');
var osm = new
  L.TileLayer('https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png', {
    attribution: 'Map © <a
    href="https://www.openstreetmap.org/copyright">OpenStreetMap</a>
    contributors',
    maxZoom: 18
  });
map.addLayer(osm);

// Fetch the GeoJSON of the area and add it to the map
var mapit = mapit || {};
[...]

PR #1330, opened 2022-08-03
[#818] Add special link for MapIt areas

garethrees:
## Relevant issue(s)

Simple version of https://github.com/mysociety/whatdotheyknow-theme/issues/818.

## What does this do?

Link to MapIt area if a body is tagged with `mapit:AREA_ID`, and/or the
"parent" area if tagged with `parent_mapit:AREA_ID`.

Also links to an on-site search to find other bodies in the same area.

## Why was this needed?

Helps users figure out if a body is relevant to their interests.

## Implementation notes

Not sure about the "Parent area" phrasing – could we improve that?

## Screenshots

Screenshot 2022-08-03 at 10 57 41

### DIFF
--- a/lib/views/public_body/_more_info.html.erb
+++ b/lib/views/public_body/_more_info.html.erb
@@ -62,6 +62,18 @@
   <% end %>
 <% end %>
 
+<% if public_body.has_tag?('mapit') %>
+  <% public_body.get_tag_values('mapit').each do |tag_value| %>
+    <%= link_to _('Area covered'), "https://mapit.mysociety.org/area/#{tag_value}.html" %>
+
+    <small>
+      [<%= link_to _('others in this area'), list_public_bodies_by_tag_path("mapit:#{tag_value}") %>]
+    </small>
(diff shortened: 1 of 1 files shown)


RichardTaylor:
Alternate, or possible additional, route to linking to the parent areas shown at

mdeuk:
> Using categories we can get eg.
> 
> > Abbey Cwm-hir Community Council
> 
> > [A town or parish council](https://www.whatdotheyknow.com/body/list/parish_council) and [a body in the Powys Council area](https://www.whatdotheyknow.com/body/list/parent_mapit:2595) ([admin](https://www.whatdotheyknow.com/admin/bodies/139694))

I **like** this idea. 👍

garethrees:
I've removed the `parent_area` portion of this PR, as it's a bit confusing and we're considering a different approach, so that we can get the main mapit link merged. Can circle back on `parent_area` another time.

Screenshot 2022-08-10 at 20 16 33

Issue #1345, opened 2022-08-10
"Body" or "authority"?


WilliamWDTK:
I suspect for Alaveteli itself, they'll always be bodies, but otherwise, what are they?

It doesn't quite seem right to me to call a school or a hospital an "authority".  How are they referred to in law?

confirmordeny:
In the Freedom of Information Act 2000, the term "public authority" is used. For example, in section 1: "Any person making a request for information to a public authority is entitled [...]". A public authority can be natural person such as a GP. The term body suggests that the entity is incorporated as a separate legal person. EIR also uses "public authority".

I believe that we use "body" in the code but "public authority" everywhere else.

RichardTaylor:
We have "body" in URLs but "authority" in text. 

"Organisation" is another term we use. 

Note we list some companies in respect of eg. being a pharmacy. 

Do we have text on the site where this is a problem?

RichardTaylor:
Change eg. 

>Found 25669 public authorities in the category ‘Schools’

to

>Found 25669 organisations in the category ‘Schools’

confirmordeny:
I think "public authority" is better in the case of privately owned businesses performing public functions. For example, I would prefer call Boots a public authority than a public body.

WilliamWDTK:
Calling Boots an authority sounds odd to me, really, since it isn't what one would call an "authority".  I think the main problem here is that there is not one word which covers all `bodies`.  "Body" is perhaps the most neutral, but it's still not accessible, and it was accessibility/comprehensibility that I was aiming for.

WilliamWDTK:
@garethrees did mention on the call that he might look back at when/why it was changed with regard to the mismatch between code and general public text.

PR #1350, opened 2022-08-14
Update how.html.erb

confirmordeny:
Changed "bodies" to "authorities". I think we are supposed to use "body" in the code and "authority" everywhere else.

## Relevant issue(s)
I think we are supposed to use "body" in the code and "authority" everywhere else.

## What does this do?
Changed "bodies" to "authorities". 

## Why was this needed?
Consistency.

### DIFF
--- a/lib/views/help/how.html.erb
+++ b/lib/views/help/how.html.erb
@@ -136,7 +136,7 @@
     <a href="#accidentally_released">#</a>
   </h4>
   <p>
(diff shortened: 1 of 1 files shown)


garethrees:
Related to https://github.com/mysociety/whatdotheyknow-theme/issues/1345.

garethrees (approved):

PR #1352, opened 2022-08-14
Update requesting.html.erb

confirmordeny:
## Relevant issue(s)
There is a UK charity that helps with whistle blowing. The UK is no longer in the EU so I have reworded the part re the EU Human Rights Defenders Relocation Platform. I didn't remove it because lots of EU citizens live in the UK and it is quite easy to travel between Northern Ireland and the Republic of Ireland.

## What does this do?
Improves the help text.

## Why was this needed?
See above.

## Implementation notes
See above.

## Screenshots
Not needed.

## Notes to reviewer
See above.

### DIFF
--- a/lib/views/help/requesting.html.erb
+++ b/lib/views/help/requesting.html.erb
@@ -713,11 +713,15 @@
       </p>
       <p>
         You may wish to consider setting up a new account for your
-        whistleblowing request as your history of requests may help people identify you.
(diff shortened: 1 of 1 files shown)


mdeuk (approved):
This is a good idea - given Brexit is now a reality we ought to ensure we link to a resource that our user base would be able to use, so having options for GB and UK(NI) / EU is a good thing to do. 🇪🇺

PR #1376, opened 2022-08-28
help text - info may be out of date

confirmordeny:
Information may be out of date on the day it is published

## Relevant issue(s)
Information may be out of date on the day it is published. This is because FOI is  aright to information held by public authorities that may or may not be up to date.

## What does this do?
Improves help text. Provides clarity to users.

## Why was this needed?
See above.

## Implementation notes
See above.

## Screenshots
N/A

## Notes to reviewer
See also: https://wdtkwiki.mysociety.org/wiki/Draft_legal_notice_for_the_site

### DIFF
--- a/lib/views/help/how.html.erb
+++ b/lib/views/help/how.html.erb
@@ -21,8 +21,8 @@
     Some public bodies will release information that is many years out of date on the site. 
     This can happen if the requester asks for information relating to events that happened a number of years ago. 
     This can also happen if a document has not been updated for some time, and the requester asks for the latest version. 
-    This means that some of the information that is published on the site will already be out of date on the day it is published.
-     We never remove material from the site just because it is old or because a newer version of the material exists. 
+   This means that some of the information that is published on the site will already be out of date on the day it is published.
+   We never remove material from the site just because it is old or because a newer version of the material exists. 
     This is because we are trying to create a permanent archive.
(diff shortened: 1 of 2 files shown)


lib/views/help/how.html.erb
```diff
@@ -12,6 +12,18 @@
...
+   <h2 id="information_may_be_out_of_date">
+    Information you find on the site may be out of date
+    <a href="#information_may_be_out_of_date">#</a>
+  </h2>
+  <p>
+    Some public bodies will release information that is many years out of date on the site. 
+    This can happen if the requester asks for information relating to events that happened a number of years ago. 
+    This can also happen if a document has not been updated for some time, and the requester asks for the latest version. 
+    This means that some of the information that is published on the site will already be out of date on the day it is published.
+     We never remove material from the site just because it is old or because a newer version of the material exists.
```
mdeuk, line 25: I think there is a double space at the start of line 25

 ```suggestion
    This means that some of the information that is published on the site will already be out of date on the day it is published.
    We never remove material from the site just because it is old or because a newer version of the material exists. 
```

mdeuk (approved):
Looks great - one possible glitch, but honestly, I think this is good to go.

mdeuk (approved):
👍

PR #1386, opened 2022-09-03
Update requesting.html.erb - no brackets for year of an Act of Parliament

confirmordeny:
We don't need to put the year in brackets in the name of an Act of Parliament. I know that the NHS page we link to does but I don't think that's right. See also: https://www.legislation.gov.uk/ukpga/1990/23/contents

### DIFF
--- a/lib/views/help/requesting.html.erb
+++ b/lib/views/help/requesting.html.erb
@@ -557,7 +557,7 @@
         such as the
         <a href="https://www.nhs.uk/common-health-questions/nhs-services-and-treatments/can-i-access-the-medical-records-health-records-of-someone-who-has-died/"
         title="NHS England guidance on the Access to Health Records Act 1990">
-        Access to Health Records Act (1990)</a> and the
+        Access to Health Records Act 1990</a> and the
         <a href="https://www.gov.uk/guidance/request-records-of-deceased-service-personnel"
         title="MoD Guidance: Request records of deceased service personnel">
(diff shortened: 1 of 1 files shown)


mdeuk (approved):
Agreed, this seems to be more fitting with established practice.

I took a moment to look up what The National Archives give as advice - [they suggest](https://www.legislation.gov.uk/help#referLegislation) the following:

> **Public General Acts of the UK Parliament**
> 
> These may be cited by the short title (which includes the year) and chapter number (bracketed), e.g. Constitutional Reform Act 2005 (c. 4).
> 
> Citations of pre-1963 Acts may also contain a reference to the 'regnal year'(that is, the year of the sovereign's reign) of the session of parliament in which the Act was passed, e.g. Statute of Westminster 1931 (22 and 23 Geo. 5 c. 4). This means that the Act was passed in 1931 during the session of Parliament spanning the 22nd and 23rd years of the reign of King George the Fifth.
> 
> **Local Acts of the UK Parliament**
> 
> These may be cited by the short title (which includes the year) and chapter number in Roman numerals (bracketed), e.g. London Local Authorities Act 1996 (c. ix)

mdeuk:
A [handy guide from the University of Cumbria](https://my.cumbria.ac.uk/media/MyCumbria/Documents/Library/Quick-guide-to-CTR-Legal-Referencing.pdf) also agrees. This is, of course, a resource designed for academia - but it is helpful as a confirmatory source.

gbp (approved):

PR #1421, opened 2022-09-29
Make it clearer that people can use the site for free

confirmordeny:
## Relevant issue(s)
Make it clearer that people can use the site for free.

## What does this do?
Amends help pages to make it clearer that people can use the site for free.

## Why was this needed?
(1) At a time when the cost of living is increasing in the UK and major national newspapers are talking about a cost of living crisis, it might be helpful to make it even clearer that people can use the site for free.

(2) We are all about transparency so making it really clear at the outset that you can make FOI requests for free is in line with our principles.

(3) In behavioural economics, there is something called 'the Power of Free'. My limited understanding of this concept is that people have a very high preference for things that are free which is higher than you might otherwise expect even by comparisons to things that cost almost nothing e.g. one penny (GBP). If I have got this right then even of the people who think the cost of using the site is likely be very low and easily affordable the knowledge that it is free will encourage a higher proportion to sign up.

https://www.forbes.com/sites/piyankajain/2018/03/01/5-behavioral-economics-principles-for-marketeers/?sh=51c10c8e28eb

## Implementation notes
N/A.

## Screenshots
N/A.

## Notes to reviewer
I am not an experienced coder.

### DIFF
--- a/lib/views/help/about.html.erb
+++ b/lib/views/help/about.html.erb
@@ -31,6 +31,15 @@
       received, are automatically published on the website for you and anyone
       else to find and read.
     </dd>
+     <dt id="does_it_cost_money">Does it cost money to get an account on the site? <a href="#does_it_cost_money">#</a> </dt>
+    <dd>
+      <p>No. You can create an account on the site for free. Once you have an account, 
+      you can make up to ten freedom of information requests free of charge each day. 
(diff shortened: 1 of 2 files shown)


confirmordeny:
See also: https://github.com/mysociety/whatdotheyknow-theme/issues/1422 which this change would help to address.

gbp (approved):

PR #1504, opened 2022-11-09
[#1503] ICO appeal link outdated on `/help/unhappy`

### LINKED ISSUE #1503: Link to ICO in `/help/unhappy` outdated
The link that we use in [`/help/unhappy#refer-to-ico`](https://www.whatdotheyknow.com/help/unhappy#refer-to-ico) is out-of-date. Whilst it _does_ go to a working page on the ICO website, the link from there is broken.

We should update the link we use to the revised link:

Old: https://ico.org.uk/make-a-complaint/foi-and-eir-complaints/
New: https://ico.org.uk/make-a-complaint/foi-and-eir-complaints/foi-and-eir-complaints/

mdeuk:
## Relevant issue(s)

Fixes #1503

## What does this do?

This fixes the outdated link to the ICO, and also adds their email address, since we are advising users that they can email §50 concerns to the ICO… the address is well hidden on their website!

A slight change is also made to the OSIC link, as there is a more direct page we can usefully direct users toward.

## Why was this needed?

ICO website has been updated and the link we are using doesn't degrade gracefully.

## Implementation notes

Nothing of note

## Screenshots

N/A

## Notes to reviewer

Nothing explicit. I did ponder if we should obfuscate the email address, but in reality, it is published in so many locations that I doubt it would have any practical effect.

### DIFF
--- a/lib/views/help/unhappy.html.erb
+++ b/lib/views/help/unhappy.html.erb
@@ -122,7 +122,9 @@
   <p>
     While the Information Commissioner asks for their form to be completed and
     evidence supporting a referral to be attached in practice we are aware they
-    routinely consider referrals made via a simple email setting out the
+    routinely consider referrals made via a simple email, to <strong><code>
+    <a href="mailto:ICOCasework@ico.org.uk?subject=FOI%2FEIR%20Complaint">
+    icocasework@ico.org.uk</a></code></strong>, setting out the
     issues, containing a link to the request on WhatDoTheyKnow.com to provide
     evidence of the full history of the correspondence related to the request.
     If you do wish to download copies of correspondence, and attachments, to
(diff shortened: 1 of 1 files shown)


FOIMonkey:
Thanks for this. We've had a user contact us to point out the form error to us today, so it'd be good to get this done.

lib/views/help/unhappy.html.erb
```diff
@@ -111,7 +111,7 @@
 
   <p>
     To make a referral,
-    start by reading <a href="https://ico.org.uk/make-a-complaint/foi-and-eir-complaints/">the
```
garethrees, line 114: This URL (now?) works, but the addition saves a redirect 👍

garethrees (approved):
LGTM! 👍

lib/views/help/unhappy.html.erb
mdeuk: Looks like they fixed it - the page was doing something funky until recently! 😅

PR #1523, opened 2022-11-26
fixing a link in the glossary

confirmordeny:
fixing a link in the glossary

## Relevant issue(s)
See above.

## What does this do?
See above.

## Why was this needed?
See above.

## Implementation notes
N/A.

## Screenshots
N/A.

## Notes to reviewer
N/A.

### DIFF
--- a/lib/views/help/glossary.html.erb
+++ b/lib/views/help/glossary.html.erb
@@ -1,0 +1,302 @@
+<% @title = "Freedom of information glossary" %>
+<%= render :partial => 'sidebar' %>
+<div id="left_column_flip" class="left_column_flip">
+ <h1 id="top">Freedom of information glossary</h1>
+ <p>This glossary provides guidance on some of the abbreviations and special terms used by <a href="#foi_officer">FOI officers</a> and transparency campaigners. This glossary also explains some of the terms used in our help pages. </p>
+<br/>
+ <p>
+<a href="#letter_A">A</a>
+<a href="#letter_B">B</a>
+<a href="#letter_C">C</a>
+<a href="#letter_D">D</a>
+<a href="#letter_E">E</a>
+<a href="#letter_F">F</a>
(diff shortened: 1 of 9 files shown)


lib/views/help/glossary.html.erb
```diff
@@ -32,13 +32,16 @@
...
 <dt id="api">API</dt>
 <dd>API stands for Application Programming Interface. An API is a piece of software that allows two computer programmes to exchange information. See also: <a href="https://www.whatdotheyknow.com/help/api">https://www.whatdotheyknow.com/help/api</a> </dd>
 <dt id="">Applicant blind</dt>
-<dd>The principle that the information released in response to an <a href="#eir">EIR</a> or <a href="#foi">FOI</a> request should be the same no matter who the <a href="#requester">requester</a> is. In practice, <a href="#public_authorities">public authorities</a> can take the identity of the requester into account in certain situations e.g. when applying the <a href="#cost_limit">cost limit</a>.</dd>
+<dd>The principle that the information released in response to an <a href="#eir">EIR</a> or <a href="#foi">FOI</a> request should be the same no matter who the <a href="#requester">requester</a> is. In practice, <a href="#
+ 
+ 
+ orities">public authorities</a> can take the identity of the requester into account in certain situations e.g. when applying the <a href="#cost_limit">cost limit</a>.</dd>
```
garethrees, line 44: Doesn't look like this should have newlines breaking the href?

garethrees (changes requested):
e3d4b92edcb099399add100ea7852fe71e1c9c9b looks good in general, but looks like it may still have one broken link

lib/views/help/glossary.html.erb
garethrees: Also, think it needs the same change from `#public_authorities` to `#public_authority`.

gbp (approved):

PR #1525, opened 2022-11-26
Don't make assumptions re users identity

confirmordeny:
We do not check users identity. FOI officers should not assume the requester is the same person as someone they have dealt with previously even if both have the same name.

## Relevant issue(s)
Ensure public authorities don't wrong infer users identity and then respond in the wrong way e.g. by copying in another email address, adding in someones middle name, qualifications etc.

## What does this do?
See above.

## Why was this needed?
See above.

## Implementation notes
N/A.

## Screenshots
N/A.

## Notes to reviewer
N/A.

### DIFF
--- a/lib/views/help/officers.html.erb
+++ b/lib/views/help/officers.html.erb
@@ -128,9 +128,9 @@
         </li>
       </ul>
     </dd>
-      <dt id="identity_checks">
-      Do you check the identity of your users?
-      <a href="#identity_checks">#</a>
+      <dt id="identity_of_requester">
+      How can we be sure the requester is the same person we have dealt with previously?
+      <a href="#identity_of_requester">#</a>
     </dt>
     <dd>
       <p>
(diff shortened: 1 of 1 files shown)


garethrees (commented):
Nice in principle, but I do wonder whether the phrasing is advantageous to us. Does it sound irresponsible to an uncharitable reader? I think the main point we're trying to make is one request from the name "Bob Smith" may or may not be the same "Bob Smith" as the next one. I don't think that's about our identity checks so much.

I don't have a better suggestion this minute – I'll mull it over – but thought I'd add these notes now in case someone thinks of something (or indeed, thinks the proposal is totally fine)

confirmordeny:
I have tried to address Gareth's feedback. I have simplified the wording here to say:
"We would advise public authorities not to assume that a person writing to them via our site is the same as a person they have dealt with previously under the same name. "

garethrees (approved):

garethrees:
Thanks! Merged as 2ba0f43.

Issue #1434, opened 2022-10-04
Present FOI & Transparency TICTeC videos on the front page


garethrees:
A specific of https://github.com/mysociety/whatdotheyknow-theme/issues/791.

Basically copy & paste of https://economicprinciples.org

<img width="1576" alt="Screenshot 2022-10-04 at 17 28 03" src="https://user-images.githubusercontent.com/282788/193874243-129d8e77-de76-446d-905e-38e52b2c2c95.png">

It shows that we're about more than just running a service. Not worth a huge engineering effort to automate as they're relatively low frequency.

garethrees:
> Do we have existing content relevant to the WDTK readership?

Probably

> We could create specific content

This issue isn't about this – it's about better promoting what we have right now, and if some specific is created later, having an established place to put it.

> Are we just boosting viewcounts and donating Google juice to other parts of the organisation?

As explained above, mySociety does not only run services; it attempts to fix underlying societal issues. The two go hand in hand. We are not in competition with ourselves.

garethrees:
> Do we have existing content relevant to the WDTK readership?

From a quick scan:

* [Workshop: What can you do with FOI using WhatDoTheyKnow?](https://www.youtube.com/watch?v=meaLp7p2Yok)
* [Reforming Freedom of Information: mySociety policy paper launch event](https://www.youtube.com/watch?v=AbXowwJDn3M)
* [The FOI Clearing House: an openDemocracy investigation into FOI at the heart of government](https://www.youtube.com/watch?v=wXbhBHfeOzk) extracted from [TICTeC Show & Tell - Right to Know across Europe](https://www.youtube.com/watch?v=Mfy0N50s8dU)
* [A change in the law for school starters in Scotland — through FOI](https://www.youtube.com/watch?v=u8wM2wJ7N4o) extracted from [TICTeC Show & Tell - Right to Know across Europe](https://www.youtube.com/watch?v=Mfy0N50s8dU)
* [Regulating Access to Information](https://www.youtube.com/watch?v=aDfh0IH4K3U)
* [TICTeC Civic Tech Surgery #3: Accessing quality information](https://www.youtube.com/watch?v=fVrAYctN7sY) – don't know how much FOI is in this one
* [Source & manage Freedom of Information data for impactful investigations & campaigns with Alaveteli](https://www.youtube.com/watch?v=n7CUaEdbRM0) – a bit more general FOI & Alaveteli
[...]

Issue #791, opened 2021-03-16
Refresh and improve the home page / front page


RichardTaylor:
The headline is currently:
>Get answers from the government and public sector

how about:

>Get government and public sector information published

Can we do more to stress the value of having a FOI response published - to support whatever anyone does with it, eg. journalism, lobbying, research etc. ?

The front page states:
> It also publishes all requests online.

rather than eg. "all requests and any responses are published online" - though the publication of responses is mentioned elsewhere on the page. 

There is also quite a prominent technical error:
>You have the right to request information from any publicly-funded body

FOI and similar access to information laws apply to public bodies, not publicly-funded bodies.

RichardTaylor:
Other Alaveteli installations have some much more engaging and interesting content on their front pages than WhatDoTheKnow does in the UK. 

**Ukraine's https://dostup.pravda.com.ua/**

There's section titled: "The lawyer comments" « Expert annotations. We could copy this and highlight advice and commentary on requests from the WhatDoTheyKnow.com team. Related ticket https://github.com/mysociety/alaveteli/issues/34 but perhaps we could work with what we currently have - for example starting a WhatDoTheyKnow-Comments User?

There are graphically engaging "publications" and "projects" sections. Perhaps these are "just" blog article links - but the integration, and particularly the images, are excellent.  WhatDoTheyKnow articles from the mySociety blog aren't shown at all on the WhatDoTheyKnow.com front page, there is a link to an integration with the mySociety blog, but it is unsophisticated and eg. it isn't possible to link to individual articles. 

**Ask the EU https://www.asktheeu.org**
Again far better blog integration. 


Also https://transparencia.be and https://www.asktheeu.org have a video on their front pages. I don't personally think that's the way to go. 

I think almost all the other instances of Alavetli have a more inviting and modern looking front page than WhatDoTheyKnow - from my looking though https://alaveteli.org/deployments/ 

The image on WhatDoTheyKnow's front page is really quite dismal - a brighter image with a background colour over the top for text as used elsewhere might be preferable to a drab image.

RichardTaylor:
We don't know why some installations of Alaveteli thrive and others don't. 

I suspect that it's probably more down to local laws/culture than the design of a site's front page, however it may be that if a site presents itself as being "about" high-impact journalism and campaigning that might deter more low-level and personally/locally focused use of the service.

WilliamWDTK:
[...]

PR #1541, opened 2022-12-22
[#1434] Added video section in Homepage

lucascumsille:
Fixes: https://github.com/mysociety/whatdotheyknow-theme/issues/1434
A specific of https://github.com/mysociety/whatdotheyknow-theme/issues/791.

@garethrees 
Before merging a couple of things:

- In this PR, I have added only three videos, but let me know if you want to add the rest of the list. My only concern is that some of the titles differ in length. The card layout takes that into account and makes all the cards have the same height, but of course, some of them will have more extra white space between the inner elements.
- Is the position alright? Right below the how it works section.
Screenshot 2022-12-23 at 08 50 47

I thought this would be a better place because, in terms of info and content, we haven't reached the page section that has more content to read; therefore, I thought users would probably have the chance to see this content in comparison to leaving it at the end of the page.

- I left a comment of for the [donation button here](https://github.com/mysociety/whatdotheyknow-theme/issues/1434#issuecomment-1360998309)
- In terms of styling I thought of these two options, let me know which one you prefer. Personally I find the card layout easier to scan:

### Option 1
Screenshot 2022-12-23 at 08 50 37

### Option 1 mobile version
Screenshot 2022-12-23 at 08 44 33

In this one we could replace the link for a button styling instead, if you are not too keen on the links with icons. 
What I like about them, without looking at the label they already telling you what they are.

### Option 2
Screenshot 2022-12-21 at 11 17 41

Let me know your thought and I'll fix the PR


lucascumsille:
@garethrees One thing it would be good to have in mind. Do you want me to change the background colour of the video section, so it doesn't look like the "Recent Videos" section and the "Who can I request information from?" are from the same section. I couldn't find many examples on the current website to use, at the moment we have `#f3f1eb` and `#ffffff`, but these ones are above and below the video section so potentially I'll need to create a new background colour to make it differ from those sections.

garethrees (changes requested):
This looks great!

> I have added only three videos

This is plenty.

> My only concern is that some of the titles differ in length…

I think it looks great with the current videos and I'm not _too_ concerned that we'll have titles that are wildly different in length.

> In terms of styling I thought of these two options, let me know which one you prefer. Personally I find the card layout easier to scan

I agree, the card layout is lovely. Could we use YouTube embeds, though? Only having to add a new URL to the `@research_videos` data structure would be much easier, rather than having to copy/paste the titles and create a thumbnail image every time we want to add a video. Keeping the cost of change low is very valuable.

> I left a comment of for the [donation button here](https://github.com/mysociety/whatdotheyknow-theme/issues/1434#issuecomment-1360998309)

I'd like the donation button CTA as described, since that's one of the main reasons for adding these sections. I agree we should still have a "browse more" link that goes to our channel, but as the secondary CTA. I'm happy with the phrasing as I wrote it in the issue for now; this isn't worth lots of people's time this minute. We can tune it later on.

> Is the position alright? Right below the how it works section. I thought this would be a better place because, in terms of info and content, we haven't reached the page section that has more content to read; therefore, I thought users would probably have the chance to see this content in comparison to leaving it at the end of the page.

I'm not sure what you mean by "we haven't reached the page section that has more content to read", but as with [the research papers](https://github.com/mysociety/whatdotheyknow-theme/pull/1542#pullrequestreview-1235591878), ideally we'd position this below the "Who can I request information from?" / "What information has been requested?" section. The reason for this is that the research papers and videos are more incidental content, whereas the Alaveteli-generated lists are more directly about what to do _on the site_.

I can see that we don't have a great way of doing that at this point, so I've made a [pull request](https://github.com/mysociety/alaveteli/pull/7492) – which will hopefully be merged by the time you read this! – that makes it easier to do this and keep the overrides a little simpler.

Should be an easy change:

* Create `lib/views/general/_frontpage_extra.html.erb`
* Add `<%= render :partial => "frontpage_videos" %>` to that file

lucascumsille:
I'm glad you like it @garethrees 

> I agree, the card layout is lovely. Could we use YouTube embeds, though? Only having to add a new URL to the @research_videos data structure would be much easier, rather than having to copy/paste the titles and create a thumbnail image every time we want to add a video. Keeping the cost of change low is very valuable.

This is how the embeds would look like with the provided space:
Screenshot 2023-01-09 at 09 53 03

Without the YouTube API I don't think we can get the title. 

Instead of using the embed we could use the "ID" of the video to pull the thumbnail and the link. But the title would be needed to be added manually. This way we could keep the same card aesthetic without using the API. For the image Alt I would use the title.
Example:
`<% 
    @research_videos = [
    { title: 'Reforming Freedom of Information: mySociety policy paper launch event',
        id: 'AbXowwJDn3M'},
    { title: 'The FOI Clearing House: an openDemocracy investigation into FOI at the heart of government',
        id: 'wXbhBHfeOzk',}
    { title: 'Workshop: What can you do with FOI?',
        id: 'meaLp7p2Yok'},
    ]
%>`

In terms of the donation would this work for you? Basically gives more priority the donation action than the "Browse more", but if you want to make the "Browse more" more obvious I could make it a secondary button, but I don't think is that advisable if the main purpose is for people to donate.

Screenshot 2023-01-09 at 10 27 56

garethrees:
> Instead of using the embed we could use the "ID" of the video to pull the thumbnail and the link. But the title would be needed to be added manually. This way we could keep the same card aesthetic without using the API. For the image Alt I would use the title.

This looks great – the less we have to do to add a video to the list the better.

> In terms of the donation would this work for you?

Here's what I had in mind:

Screenshot 2023-01-09 at 10 51 35

I think it's important to indicate _why_ we're asking for donations. The reason is that it supports us do the interesting things that we're showing immediately above.

lucascumsille:
@garethrees I just pushed three commits that:

- [ ] Leaves the donation button with this layout 
211291822-18e2d6b7-ff0f-4944-95a2-0c441133f1fd

- [ ] The ID now replaces the url and img. The title will also be the alt for the image.
- [ ] `Create lib/views/general/_frontpage_extra.html.erb
Add <%= render :partial => "frontpage_videos" %> to that file`

Let me know if there is anything else you'd like me to do =)

lucascumsille:
Thanks for the feedback @garethrees I have rebased and fixed the indention on this PR.
Let me know if there is anything else.

lib/views/general/_frontpage_videos.html.erb
```diff
@@ -0,0 +1,33 @@
...
+
+<div class="homepage-video-section" id="homepage-video">
+    <div class="row homepage-video-section__vs-content">
+        <div class="vs-content__headings ">
+            <h2><%= _("Recent videos") %></h2>
+        </div>
+        <div class="vs-content__video-grid col-3-grid">
+            <% @research_videos.sample(3).each do |article| %>
+                <div class="video-grid__item">
+                <img class="item-content__img" src="http://img.youtube.com/vi/<%= article[:id] %>/hqdefault.jpg" alt="<%= article[:title] %>" loading="lazy">
```
garethrees, line 20: Should this be a http**s** url? No need to fix – I can fix it in the merge, just need to know whether to tweak :)

lucascumsille: Hi Gareth, yes please I didn't see that. It should be https.

garethrees:
I'm getting having doubts about including the donation value on the button so will mull that over, but aside from that this is ready to go. We'll roll out in conjunction with https://github.com/mysociety/whatdotheyknow-theme/pull/1146 and https://github.com/mysociety/Design-Internal-tasks/issues/8#issuecomment-1410057767.

lib/views/general/_frontpage_videos.html.erb
```diff
@@ -42,7 +42,7 @@
     </div>
 
     <div class="cta-button-wrapper">
-      <a class="button" href="https://www.mysociety.org/donate/?utm_medium=link&how-much=10">Donate £10</a>
+      <a class="button" href="<%= donation_url('how-much' => 10, utm_content: 'homepage donate 10', utm_campaign: 'wdtk-homepage-videos') %>">Donate £10</a>
```
garethrees, line 45: @ajparsons just want to check the utm tracking here.

Currently this generates the following URL (split for ease of reading):

```
https://www.mysociety.org/donate?
  how-much=10&
  utm_campaign=wdtk-homepage-videos&
  utm_content=homepage+donate+10&
  utm_medium=link&
  utm_source=whatdotheyknow.com
```

I think I'll drop the £ value and change the button text to "Donate now", so that will end up as:

```
https://www.mysociety.org/donate?
  how-much=10&
  utm_campaign=wdtk-homepage-videos&
  utm_content=homepage+donate+now&
  utm_medium=link&
  utm_source=whatdotheyknow.com
```

Is this what you'd expect? Easy to change if there's a better combo.

ajparsons: Good on campaigns settings, if dropping "donate £10" drop the how-much because our suggested off-on donation is now lower than that. If keeping it, also add an 'how-often=one-off' parameter.

See https://github.com/mysociety/orgsites/pull/1321 for more info on that.

Can be used in combination for a double button like this (example from the blog)

image

PR #1556, opened 2023-01-12
Remove custom autoloaded path

gbp:
## Relevant issue(s)

https://github.com/mysociety/alaveteli/issues/6341

## What does this do?

Remove custom autoloaded path

## Why was this needed?

Fixes issues when using theme on Rails 7 as this was causing the post install script to error with:
```
Zeitwerk::NameError: expected file ./lib/censor_rules.rb to define constant CensorRules, but didn't
```

Needed to explicitly require some files as Rails convention is to not autoload files under `lib/`.

See: https://github.com/rails/rails/issues/38671#issuecomment-947449850

### DIFF
--- a/lib/alavetelitheme.rb
+++ b/lib/alavetelitheme.rb
@@ -31,15 +31,6 @@ Rails.application.config.assets.precompile.unshift(LOOSE_THEME_ASSETS)
 
 Rails.application.config.assets.precompile << ["tests.js"]
 
-# In order to have the theme lib/ folder ahead of the main app one,
-# inspired in Ruby Guides explanation: http://guides.rubyonrails.org/plugins.html
-%w{ . }.each do |dir|
-  path = File.join(File.dirname(__FILE__), dir)
-  $LOAD_PATH.insert(0, path)
-  ActiveSupport::Dependencies.autoload_paths << path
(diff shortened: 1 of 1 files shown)


garethrees (commented):
Looks fine.

Worth just checking that files in `theme/lib` continue to be loaded in the same order as they were with the removed code.

gbp:
> Worth just checking that files in `theme/lib` continue to be loaded in the same order as they were with the removed code.

This is set in core https://github.com/mysociety/alaveteli/blob/91d9f384cd7051897af8e71c76e1b077e023cc6f/config/initializers/theme_loader.rb#L12

```ruby
[2] pry(main)> $LOAD_PATH
=> ["./alaveteli/lib/themes/whatdotheyknow-theme/lib",
 "./alaveteli/vendor",
 "./alaveteli/app/controllers/concerns",
 "./work/alaveteli/app/models/concerns",
 "./work/alaveteli/app/controllers",
 "./work/alaveteli/app/helpers",
 "./work/alaveteli/app/mailers",
 "./work/alaveteli/app/models",
 "./work/alaveteli/app/services",
 "./work/alaveteli/app/validators",
...
```

Issue #1473, opened 2022-10-27
Investigate why the volunteer contact form is generating so much spam


sallytay:
Since introducing the new volunteer contact form (https://github.com/mysociety/whatdotheyknow-theme/issues/1179) it has been noted that it generates a lot more spam that the general contact form

For example there have been 33 this month, 12 this week.

It is not clear why this is happening - could we investigate this to see whether there is anything we can do to prevent this?

FOIMonkey:
We've had 16 this week, and 44 in the past month. It seems to be getting worse, and there are consistently several messages a day that need dealing with.

mdeuk:
> We've had 16 this week, and 44 in the past month. It seems to be getting worse, and there are consistently several messages a day that need dealing with.

The new "Inbox Monster" alpha now has some capability for handling these messages - a one-click "form spam" option which takes corrective action on the thread itself, and then logs it into a [spreadsheet](https://docs.google.com/spreadsheets/u/3/d/e/2PACX-1vTzhtZBOvS_8iVtEbrI5ls3xosDgc-FpCyO3jKpY8qQWXatMSpjn_Ep6GTVs5Sv2maX8uW-QLIDYPf_/pubhtml).

There's also some bad jokes within! 😇 

Related:
- #1272

HelenWDTK:
+1 Anecdotally, the spam seems to be getting worse, and the inbox monster data supports this. It'd be good to get the same kind of spam protection on the volunteer form as the main contact one if possible.

HelenWDTK:
22% of all new threads in the inbox over the past 7 days have been spam received via the volunteer form. The problem is continuing to escalate.

garethrees:
The honeypot form field definitely works as expected:

```diff
diff --git a/app/controllers/help_controller.rb b/app/controllers/help_controller.rb
index 4ddfa8caf..fc6e45f78 100644
--- a/app/controllers/help_controller.rb
+++ b/app/controllers/help_controller.rb
@@ -86,6 +86,7 @@ def contact_params
   end

   def catch_spam
+    puts "CATCHING SPAM: #{params[:contact]}"
     return unless request.post? && params[:contact]
     return if params[:contact].fetch(:comment, '').blank?
```

When the `contact` parameter is submitted the controller halts processing via `catch_spam` and redirects to the homepage:

```
alaveteli-app-1  | Processing by HelpController#contact as HTML
[...]

PR #1577, opened 2023-02-01
Improve volunteer form spam detection

gbp:
## Relevant issue(s)

Connected to #1473

## What does this do?

Improve volunteer form spam detection

## Why was this needed?

We're getting lots of spam on the volunteer form. Most of the messages have the same text in `why`, `experience` and `anything_else` fields. Lets detect this and fake the message sent flash so spam bots don't change their behaviour.


garethrees (approved):
Haven't run but agree that it's a good approach and worth trying.

PR #1604, opened 2023-02-24
Add Met Police no-reply address to `model_patches.rb`

### LINKED ISSUE #1603: Add email address for Metropolitan Police to `model_patches.rb`
The Met have recently started issuing replies from `MPSdataoffice-IRU-DONOTREPLY-at-met.police.uk`, with a note in the email body that replies should go to a different address.

Since users cannot see that email address, and may not realise that a reply might not go through, we should include this address in the `invalid_reply_addresses` list used by `ReplyToAddressValidator`, to help solve the problem.

mdeuk:
## Relevant issue(s)

Fixes #1603

## What does this do?

Adds a new email address to the `invalid_reply_addresses` list used by `ReplyToAddressValidator`.

## Why was this needed?

The Metropolitan Police are now replying to some requests from a no-reply mailbox.

## Implementation notes

Nothing of note.

## Screenshots

N/A

## Notes to reviewer

Nothing of note here, it's a simple enough change. 😊

### DIFF
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -154,6 +154,7 @@ Rails.configuration.to_prepare do
     website@digital.sthelens.gov.uk
     noreply@m.onetrust.com
     no-reply@notify.microsoft.com
+    MPSdataoffice-IRU-DONOTREPLY@met.police.uk
   )
 
   User.class_eval do


garethrees (approved):
🙏

PR #1626, opened 2023-03-16
Reconcile with core

gbp:
See: https://github.com/mysociety/alaveteli/pull/7645


garethrees:
Needs an update for https://github.com/mysociety/alaveteli/pull/7645/commits/a5ea2064cb08be8a7e9256a05ebfbad823354ad3

Issue #1624, opened 2023-03-12
(Comprehensively) tag exempt charities as such


WilliamWDTK:
See [https://www.gov.uk/government/publications/exempt-charities-cc23/exempt-charities#list-of-exempt-charities](https://www.gov.uk/government/publications/exempt-charities-cc23/exempt-charities#list-of-exempt-charities). I am not aware of an existing tag for this, so I'd propose `exempt_charity`. We would presumably also want to tag them as `charity`. I suppose we could use `charity:exempt`, but that might require changes to existing code (e.g. for linking to the Commission).

WilliamWDTK:
There is not a simple comprehensive list of them, mind, so a reasonable amount of research may be required.

WilliamWDTK:
We might be able to make a request to HMRC for a list of those that are registered with them as an exempt charity for tax purposes, and we could include this reference in some tag.

mdeuk:
>  I suppose we could use charity:exempt, but that might require changes to existing code (e.g. for linking to the Commission).

From [views/public_body/_more_info.html.erb](https://github.com/mysociety/whatdotheyknow-theme/blob/master/lib/views/public_body/_more_info.html.erb): 

```erb
<% if public_body.has_tag?('charity') %>
  <% public_body.get_tag_values('charity').each do |tag_value| %>
    <% if tag_value.match(/^SC/) %>
      <%= link_to _('Charity registration'),
                    "https://www.oscr.org.uk/about-charities/search-the-register/charity-details?number=#{ tag_value }" %><br>
    <% elsif tag_value.match(/^NIC/) %>
      <%= link_to _('Charity registration'),
                    "https://www.charitycommissionni.org.uk/charity-search/?regId=#{ tag_value.gsub(/^NIC/, '') }" %><br>
    <% else %>
      <%= link_to _('Charity registration'),
                    "https://register-of-charities.charitycommission.gov.uk/charity-search/-/results/page/1/delta/20/keywords/#{ tag_value }" %><br>
    <% end %>
  <% end %>
<% end %>
```

… it should be trivial to modify this slightly, to match `charity:exempt`. I assume we'd want to link to the page on gov.uk that you've noted, as this would give context.

confirmordeny:
I agree that we should tag exempt charities as such.

mdeuk:
I've tagged a link for `charity:exempt` onto #1632. This should render in the sidebar as:
[...]

PR #1632, opened 2023-03-22
[#1624, #1631] Fix link for CCNI database, add exempt charities

### LINKED ISSUE #1631: NI Charity Commission link in `_more_info.html.erb` broken
In #1267 (#1284) we added a link to [/lib/views/public_body/_more_info.html.erb](https://github.com/mysociety/whatdotheyknow-theme/blob/master/lib/views/public_body/_more_info.html.erb) - for charities tagged `charity:NIC` - unfortunately this is now broken.

## Steps to reproduce

1. Visit listing for [National Museums of Northern Ireland - #1653](https://www.whatdotheyknow.com/body/nmni)
2. Click on the "Charity registration" link, noting that it is populated as: https://www.charitycommissionni.org.uk/charity-search/?regId=103729
3. Note that the resulting page is a search box / listing of charities, rather than something specific.

## Suggested fix

Running a search for the charity suggests that replacing `charity-search` with [`charity-details`](https://www.charitycommissionni.org.uk/charity-details/?regId=103729) ought to do the trick.

mdeuk:
## Relevant issue(s)
- Fixes #1631
- relates to #1624

## What does this do?

This amends the link we use for the Charity Commission of Northern Ireland (charities tagged `charity:NIC00000`) [^1].

It also adds a link for `charity:exempt`, which points to the [guidance on gov.uk](https://www.gov.uk/government/publications/exempt-charities-cc23/exempt-charities#list-of-exempt-charities)

## Why was this needed?

The CCNI link we use is broken and directs users to a page that isn't very helpful to them.

Secondly, at the moment we don't have a way to display an indicator for _exempt charities_. This reuses existing code to do it cheaply.

## Implementation notes

Nothing of note - however, I do ponder the regex for exempt. Theoretically, we should match `charity:exempt` and `charity:exempt00000` [^1]. I think this does the trick _as is_.

## Screenshots

N/A

## Notes to reviewer

Nothing of note.

[^1]: where `00000` is the charity registration number.

### DIFF
--- a/lib/views/public_body/_more_info.html.erb
+++ b/lib/views/public_body/_more_info.html.erb
@@ -22,7 +22,7 @@
                     "https://www.charitycommissionni.org.uk/charity-details/?regId=#{ tag_value.gsub(/^NIC/, '') }" %><br>
     <% elsif tag_value.match(/^exempt/) %>
       <%= link_to _("Exempt charity - <i>what's that?</i>"),
-                    "https://www.gov.uk/government/publications/exempt-charities-cc23/exempt-charities#list-of-exempt-charities" %><br>
+                    "https://www.gov.uk/government/publications/exempt-charities-cc23/exempt-charities" %><br>
     <% else %>
       <%= link_to _('Charity registration'),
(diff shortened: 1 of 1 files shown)


garethrees:
Nice!

Issue #1493, opened 2022-11-08
Provide a good explanation of what Environmental Information is.


FOIMonkey:
This came up in relation to a support issue relating to one of the TOCs, and I'm ticketing it so I can archive the thread! 

We list a good number of bodies that are eir_only, so there is value in making sure requesters understand what information is covered. 

This could be through better notes on the authority pages, improving the help pages or a blog post that we could link to.

confirmordeny:
See also: https://www.whatdotheyknow.com/help/glossary#environmental_information

PR #1633, opened 2023-03-22
1493 provide a good explanation of what environmental information is

HelenWDTK:
## Relevant issue(s)
#1493 
## What does this do?
Adds a help page on Environmental Information that can be expanded or improved later. 
## Why was this needed?
EIR is awesome, so we really should tell people about it. Also #IssueZero
## Screenshots
Screenshot 2023-03-22 at 22 29 50
Screenshot 2023-03-22 at 22 30 02
Screenshot 2023-03-22 at 22 28 56
Screenshot 2023-03-22 at 22 28 30
Screenshot 2023-03-22 at 22 30 20
## Notes to reviewer
I wrote this whilst I was bored waiting for my scripts to run this afternoon. It was tested via a quite heavily modified version of the WDTK theme, so perhaps check to make sure nothing odd has slipped through...

### DIFF
--- a/lib/config/wdtk-routes.rb
+++ b/lib/config/wdtk-routes.rb
@@ -38,4 +38,7 @@ Rails.application.routes.draw do
 
   get '/help/glossary' => 'help#glossary',
       as: :help_glossary
+  
+  get '/help/environmental_information' => 'help#environmental_information',
+      as: :help_environmental_information
 end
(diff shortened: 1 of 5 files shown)


lib/config/wdtk-routes.rb
```diff
@@ -38,4 +38,7 @@
 
   get '/help/glossary' => 'help#glossary',
       as: :help_glossary
+  
+   get '/help/environmental_information' => 'help#environmental_information',
```
mdeuk, line 42:
```suggestion
  get '/help/environmental_information' => 'help#environmental_information',
```

Looks like a spurious space has snuck in at the start here.

garethrees (approved):
🌳

PR #1644, opened 2023-03-31
[#541] Fix donation link

### LINKED ISSUE #541: Success donate link does not have a separate campaign tag
Success link to donate page gets marked up as the same 'donation_drive_2016' as the requests page - this should have it's own 'tagged_success' campaign tag so these can be tracked separately. 

https://github.com/mysociety/whatdotheyknow-theme/blob/master/lib/views/request/describe_notices/_successful.html.erb

garethrees:
Improve tracking of successful request donations 

Fixes #541

Screenshot 2023-03-31 at 17 44 40

Generates:

```
https://www.mysociety.org/donate?
  utm_campaign=successful_request&
  utm_content=successful+request+donate+now&
  utm_medium=link&
  utm_source=whatdotheyknow.com
```

### DIFF
--- a/lib/helpers/donation_helper.rb
+++ b/lib/helpers/donation_helper.rb
@@ -16,7 +16,6 @@ module DonationHelper
   def donation_url(options = {})
     AlaveteliConfiguration::donation_url + "?" << options.reverse_merge(
       :utm_source => 'whatdotheyknow.com',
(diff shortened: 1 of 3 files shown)


ajparsons (approved):
Super, good to break this one out (I think we want stronger messaging here in the long run - but yes)

Issue #1658, opened 2023-04-18
Update the request count on the FOI officer help page


HelenWDTK:
We've had almost half a million more requests than we mention.

PR #1659, opened 2023-04-18
Update the request count in the FOI officer help page

HelenWDTK:
Update the request count in the FOI officer help page
## Relevant issue(s)
#1658 
## What does this do?
Updates the number of requests
## Why was this needed?
The figure was very out of date
## Implementation notes
n/a
## Screenshots
n/a
## Notes to reviewer
n/a

### DIFF
--- a/lib/views/help/officers.html.erb
+++ b/lib/views/help/officers.html.erb
@@ -14,7 +14,7 @@
         mySociety.
       </p>
       <p>
-        WhatDoTheyKnow has processed over 430,000 FOI requests since its launch
+        WhatDoTheyKnow has processed over 900,000 FOI requests since its launch
         in 2008, and is the UK’s most widely-known FOI service. We also campaign
         for citizen’s rights to information and for the strengthening of FOI
         laws.


garethrees (approved):

mdeuk:
This change looks good - but just a query… we have a live(ish) count on the homepage (contained in [/lib/views/general/_frontpage_hero.html.erb](https://github.com/mysociety/whatdotheyknow-theme/blob/master/lib/views/general/_frontpage_hero.html.erb) - would that variable be accessible from help pages?

garethrees:
> would that variable be accessible from help pages 

It's not available at present, so this is the easiest change. It's possible to make it available, but not without adding complexity to the help controller (which is intended to be quite basic). The way I'd make it more widely available is a helper method. Linking to https://github.com/mysociety/alaveteli/issues/1836, but this isn't a priority right now.

mdeuk:
> > would that variable be accessible from help pages
> 
> It's not available at present, so this is the easiest change.

Ta. I agree with the change here - it's not too difficult to occasionally update it at milestones [^1] so least complex WFM. 😄 

[^1]: 1 million, 1.25 million, etc…

Issue #1663, opened 2023-04-20
Call the house rules something else


HelenWDTK:
Whilst looking at strengthening the language around unacceptable behaviour, there's been a suggestion that we change 'House rules' to 'Conditions of use'.

I'm strongly in favour of changing the name. I tend not to use house rules when talking to users, and prefer to call them 'site rules' instead.

RichardTaylor:
There is previous discussion on this point at:

https://docs.google.com/document/d/1ffX3BCFweAFH3QAWOP659HYvWYD9amdjTJ6O2uQ6ZjI/edit?skip_itp2_check=true&pli=1#bookmark=id.vnnxl9q07yi2

A concern raised was the language "House Rules" could be elitist. 

I think it's unusual language and using more common, plain English, would be preferable. 

I think:

> I recall “house rules” for social media was a fad from central government departments and similar many years ago that I think we followed.

was my contribution, see eg. 

https://gds.blog.gov.uk/social-media-house-rules/
https://www.gov.uk/government/organisations/planning-inspectorate/about/social-media-use#social-media-house-rules
https://www.google.com/search?q=site%3Agov.uk+"house+rules"

I don't think we should be following Government trends like this, doing something different would help show our independence.  


**Other Services**

* Facebook has "Community Standards" https://transparency.fb.com/en-gb/policies/community-standards 
* Reddit has the "Reddit Content Policy", which contains the site's "Rules" https://www.redditinc.com/policies/content-policy
* Twitter has "The Twitter Rules" https://help.twitter.com/en/rules-and-policies/twitter-rules
* Wikipedia has "Policies and Guidelines" https://en.wikipedia.org/wiki/Wikipedia:Policies_and_guidelines 
* There is a "Google Groups Content Policy" https://support.google.com/groups/answer/4561696?hl=en

mdeuk:
I've added a PR - #1671 - which adds a couple of different `/help` urls as a redirect for the house rules. This will allow us to refer to them using a choice of phrases, as necessary, without breaking any existing links.

garethrees:
WalesOnline uses "house rules", but that links to a "[Community Standards](https://www.walesonline.co.uk/community-standards/)" page.

![Screenshot 2023-05-12 at 10 10 23](https://github.com/mysociety/whatdotheyknow-theme/assets/282788/b76308c6-2d45-4560-a395-f04a7ae04405)

[...]

PR #1671, opened 2023-05-02
[#1663] Add additional routes for `/help/house_rules`

mdeuk:
## Relevant issue(s)

#1663

## What does this do?

This adds a new route which allows for the `/help/house_rules` page to be accessed using different paths.

## Why was this needed?

Chat on #1663, and predecessors, has raised the issue of language.

Allowing more possibilities allows us to ensure that messages to users are easy to understand, as _"house rules"_ may not mean something to many.

## Implementation notes

The basic implementation isn't anything especially complex - it is a redirect to `/help/house_rules`.

We match `:help_house_rules_resolve` against a constraint. The following are accepted values:

- `conditions_of_use`
- `site_rules`
- `terms_and_conditions`
- `terms_of_service`
- `terms_of_use`
- `the_legal_stuff`
- `the_rules`

This doesn't change the _existing_ route for `/help/house_rules`, as whilst it is trivial to include _that_ here, it is a bit less _elegant_ [^1].

## Screenshots

N/A

## Notes to reviewer

I've tested this on a sandbox install (running `0.43.0.0` / rails `7.0.4.3` / ruby `3.2.1`) and didn't encounter any problems.

[^1]: not that this is elegant… but, it prevents a slight nuisance  in regard to how we render the sidebar (e.g. not rendering a clickable 'house rules' link when we don't have to).

### DIFF
--- a/lib/config/wdtk-routes.rb
+++ b/lib/config/wdtk-routes.rb
@@ -24,6 +24,10 @@ Rails.application.routes.draw do
   get '/help/house_rules' => 'help#house_rules',
       as: :help_house_rules
 
+  get '/help/:house_rules_resolve' => redirect("/help/house_rules", status: 302),
+        as: :help_house_rules_resolve,
+        constraints: { house_rules_resolve: /conditions_of_use|site_rules|terms_and_conditions|terms_of_service|terms_of_use|the_legal_stuff|the_rules/ }
+      
   get '/help/how' => 'help#how',
       as: :help_how
 


garethrees:
Thanks for this, but I think we should instead resolve https://github.com/mysociety/whatdotheyknow-theme/issues/1663 and then update the URL. If we start using a multitude of URLs it might get quite confusing (e.g. if we send a user a URL for "conditions of use" and then they try to find the page through browsing, they won't find it.) We'd also then have to consider whether to maintain these URLs over the long term, which while small, adds a difficult-to-resolve issue and a little more code complexity.

PR #1693, opened 2023-06-01
Add North Lincs noreply address to model_patches.rb

WilliamWDTK:
https://www.whatdotheyknow.com/request/social_care_systems_and_subject_139#incoming-2323004

I'm pretty certain this is a no-reply address, but I might have got mixed up.

## Relevant issue(s)
None

### DIFF
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -160,6 +160,7 @@ Rails.configuration.to_prepare do
     Information-rights@nhsbsa.nhs.uk
     noreply-horsham@axlr8.com
     donotreply@plymouth.gov.uk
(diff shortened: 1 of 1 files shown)


mdeuk (approved):
Yep, this looks like a no-reply mailbox to me.

Curiously, the mail headers of the message being replied to suggest that it ought to have gone elsewhere…

> From: "Information Governance Team" <request-at-ig.northlincs.gov.uk>
> Reply-To: "Information Governance Team" <informationgovernanceteam-at-northlincs.gov.uk>

… yet, the Exim logs do show us using the From address. Not sure what's happened there, as I would have expected the `Reply-To` be honoured. I don't suppose it matters though.

Issue #1706, opened 2023-06-14
Add a help page on exemptions


HelenWDTK:
We should add a help page that lists the reasons why authorities can refuse FOI/FOISA requests, with links to the relevant ICO/OSIC guidance.

PR #1707, opened 2023-06-14
1706 add a help page on exemptions

HelenWDTK:
## Relevant issue(s)
#1706
## What does this do?
Adds a help page about the various FOI and FOISA exemptions, with links to guidance
## Why was this needed?
We had no way of linking users to guidance on specific issues. Help people to understand what they can and can't get released under FOI
## Screenshots
Screenshot 2023-06-14 at 14 37 44
Screenshot 2023-06-14 at 14 38 01
Screenshot 2023-06-14 at 14 38 13

### DIFF
--- a/lib/config/wdtk-routes.rb
+++ b/lib/config/wdtk-routes.rb
@@ -44,4 +44,7 @@ Rails.application.routes.draw do
   
   get '/help/environmental_information' => 'help#environmental_information',
       as: :help_environmental_information
+  
+  get '/help/exemptions' => 'help#exemptions',
+      as: :help_exemptions
 end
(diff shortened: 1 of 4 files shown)


HelenWDTK:
I've tested the links here, and everything looks good. One point of detail: I describe s36 as only having a PIT sometimes, due to the fact that a qualified person can issue a certificate. It might be that we want to amend that to say yes.

garethrees (approved):
Looks amazing!

Only suggestion would be to add a line at the top pointing people to `/help/unhappy` and our refusal advice there.

HelenWDTK:
Added the link. The wording is not the best, but borrowed from elsewhere on the help pages, so if we want to fix it, we should prob. fix both.

garethrees:
Yeah works well enough for now 👍

PR #1711, opened 2023-06-18
Links from students' union pages to parent institutions

confirmordeny:
Links from students' unions pages to the parent institution e.g. from University of Lincoln to University of Lincoln Students' Union.

## Relevant issue(s)
See above.

## What does this do?
See above.

## Why was this needed?
See above.

## Implementation notes
None.

## Screenshots
None.

## Notes to reviewer
None.

### DIFF
--- a/lib/views/public_body/_more_info.html.erb
+++ b/lib/views/public_body/_more_info.html.erb
@@ -72,6 +72,13 @@
   <% end %>
 <% end %>
 
+<% if public_body.has_tag?('su_parent') %>
+  <% public_body.get_tag_values('su_parent').each do |tag_value| %>
+      <%= link_to _('Parent institution'),
+                    "https://www.whatdotheyknow.com/body/#{ tag_value }" %><br>
+  <% end %>
+<% end %>
+
 <% if public_body.has_tag?('lei') %>
   <% public_body.get_tag_values('lei').each do |tag_value| %>
       <%= link_to _('Legal Entity Identifier'),


garethrees (approved):
Looks good.

I'm sure you're aware but just noting that this won't work with some of the current tags in the form `su_parent:PublicBody/583`

mdeuk:
> I'm sure you're aware but just noting that this won't work with some of the current tags in the form `su_parent:PublicBody/583`

Just noting, this was the [format suggested a little while back](https://mysociety.slack.com/archives/CNH83HJP4/p1683293043256249?thread_ts=1683289157.804849&cid=CNH83HJP4). I'm guessing we probably need a ticket upstream for this?

garethrees:
Ah! Copying that suggestion here for reference:

> I would actually suggest using the ID and prefixing with the model type `parent:PublicBody/2884` that way in the future maybe we could detect these and interpolate the correct name and link into a common note (edited) 

I can see the benefit in that as it would be resistant to name changes, but this PR would need extra code to handle it. Here's a sketch while I'm thinking about it, but completely untested right now:

```erb
  <% public_body.get_tag_values('su_parent').each do |tag_value| %>
       <% klass, id = tag_value.split('/') %>
       <% body = klass.constantize.find(id) %>

      <%= link_to _('Parent institution'),
                    "https://www.whatdotheyknow.com/body/#{ body.url_name }" %><br>
```

Might want to add handling for if the tag is malformed or can't find the body etc.

confirmordeny:
I went with the option that requires minimal code change. 

I agree that using ID numbers is probably better longer term.

confirmordeny:
Just to note, we don't currently expose the authority ID numbers (at least not in a clear and obvious way) which makes them less useful as a key. As an example, Wikidata uses the short name as the WhatDoTheyKnow ID as opposed to the ID number.

PR #1696, opened 2023-06-07
Add page for reporting data breaches

chrismytton:
## Relevant issue(s)

Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/1602

## What does this do?

This adds a new help page containing a form for reporting data breaches. When submitted the form is emailed to the site admins.

## Implementation notes

It follows a similar pattern to the volunteer contact form, that was added in #1208, but uses a new POST endpoint for handling the form submission.

## Screenshots

image

## Notes to reviewer

Opening this pull request for feedback and to check that this is heading in the right direction.

There's still a bit more to do on this, namely:

- [x] Write some copy for the high-level description/introduction that will go above the form
- [x] Write some copy for the **Further information** section below the form
- [x] Add some tests (wanted to check I was heading in the right direction before adding these)
- [x] Design? Think what's there is good enough to start with, but happy to hear any feedback on how it could be improved

PR #1727, opened 2023-06-28
Fix specs

chrismytton:
While doing #1696 I noticed that some of the other specs were failing, this fixes them.

I've been running the specs from within my alaveteli install with the following command (I use docker):

```
docker-compose run --rm app bin/rspec lib/themes/whatdotheyknow-theme/spec
```

Would be nice if these tests could run automatically for this repo, but guessing that would mean getting alaveteli installed and setup as part of the testing process?


gbp (approved):
Nice. Thanks for this. Really need to get some automated CI running for the theme.

PR #1734, opened 2023-07-05
Add mirror GitHub workflow

gbp:
Sync changes made on GitHub back to the mySociety git repo.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches: [master, develop]
   pull_request:
+  workflow_call:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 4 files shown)


dracos:
So every push is running CI and robocop workflows twice, once their own, and then again in order to decide whether to mirror. Given the mirror action is running the other two, can you stop their independent runs?

gbp:
> So every push is running CI and robocop workflows twice, once their own, and then again in order to decide whether to mirror. Given the mirror action is running the other two, can you stop their independent runs?

I don't believe that is the case. Unless I'm missing something, if you look at the Action tab in the Github UI, then it appears to be running only once.

dracos:
To take the PR I just opened:
* https://github.com/mysociety/whatdotheyknow-theme/actions/runs/5536311781 was the CI run
* https://github.com/mysociety/whatdotheyknow-theme/actions/runs/5536311786 was the robocop run
* https://github.com/mysociety/whatdotheyknow-theme/actions/runs/5536286421/jobs/10103745079 was the mirror CI run
* https://github.com/mysociety/whatdotheyknow-theme/actions/runs/5536286421/jobs/10103744839 was the robocop CI run
* https://github.com/mysociety/whatdotheyknow-theme/actions/runs/5536286421/jobs/10103805236 was the mirror run after those two

gbp:
Ah right I see now. I'll get that the workflows updated. Thanks for flagging.

PR #1736, opened 2023-07-05
Improve the glossary nav

garethrees:
It's a bit disorienting when you're looking for a letter and it's not next to the letter you were expecting because of the jump (e.g. I'm looking for "L", expecting it to be right of "K", but it's actually right of "I"). This PR adds missing letters, styling them with a muted colour so it's obvious there's no entries for that letter.

It also improves the styling, giving each letter a bit more of a target click space, highlights each letter as you focus (if it has entries).

Screenshot 2023-07-05 at 10 08 55

### DIFF
--- a/app/assets/stylesheets/responsive/custom.scss
+++ b/app/assets/stylesheets/responsive/custom.scss
@@ -1763,6 +1763,14 @@ dd {
   }
 }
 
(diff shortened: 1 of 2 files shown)


confirmordeny:
Looks great. Thank you.

Issue #1453, opened 2022-10-18
Overhaul the help pages, to make them shorter, prettier and more accessible.


FOIMonkey:
### What's the issue?

- ❗The current help pages are very long - in total they include 20,270 words.
- ❗There is no proper index/toc
- ❗Although we can sometimes link to individual headings, there is no easy way to jump between these.
- ❗A lot of the language we use is quite formal. 
- ❗With the exception of the Pro page, there are no images.

### Why is this a problem?

- ❌ Walls of text are intimidating for lower literacy users & second language speakers.
- ❌ If the words that we use are hard to understand, this could increase misuse
- ❌ The longer each page is, the less likely it is that anyone will read it.
- ❌ It's hard to find what you are looking for without using the browser search
- ❌ A lack of images or other elements to break up the text makes it harder to read, and visually unappealing.

### Possible solutions:

- ✅ Completely overhaul the help section and start again from the beginning. Instead of being restricted by the current structure, we could think about who the audience is that want to target with each part, or deliver the help pages by theme.
- ✅ Split the current help pages into subpages, with proper navigation links to move between them.
- ✅ Collapse the text that's not the headline/question on the longer help pages to make them seem shorter.
- ✅ Add images, or other styling to break up the wall of text.

### A more ordered International example:
🏴‍☠️ https://fragdenstaat.de/hilfe/
🏴‍☠️ https://fragdenstaat.de/en/about-us/netiquette/

Ideally I'd love to see us do this before adding yet more stuff.

FOIMonkey:
This could unlock/enable/close #1443, #1422, #1442, #1353, #1217, #1157, #1140, #1065, #1054, #1050, #1027,,#1026,#1021,
#996, #994, #991, #987, #984, #976, #952, #914, #906, #905, #878, #877, #841, #790, #787, #786, #728, #621 and many more!

WilliamWDTK:
A lot of sites have a sort of knowledgebase structure, with each page only being one or two paragraphs long, with links between them, and pages being in a category etc, and having a dedicated search facility.

RichardTaylor:
Addition to:

** Why is this a problem? **

* ❌ Our help pages are the primary place we record our policies and practices for running the service. We want it to be easy for new administrators to find, understand and apply our policies.

mdeuk:
> > Emoji
> 
> We should consider accessibility.
> 
[...]

PR #1738, opened 2023-07-05
Improve help navigation menu

garethrees:
We've got lots of help pages now, so this commit adds headings to group related pages.

There's probably lots of improvement to make, but this is a start.

A very simple version of https://github.com/mysociety/whatdotheyknow-theme/issues/1453.

Screenshot 2023-07-05 at 17 48 13


garethrees:
Here's an alternative suggestion to split the "User guides" section into two:

Screenshot 2023-07-05 at 17 54 59

WilliamWDTK:
I like the proposed split of the "User guides" section into two

garethrees:
> I like the proposed split of the "User guides" section into two

Went with this as this was the general feedback, and with the new FOI & FOISA pages tipped the balance in favour of two sections to avoid a single section looking too big.

PR #1747, opened 2023-07-12
Make correspondence headers sticky.

dracos:
## Relevant issue(s)

https://github.com/mysociety/alaveteli/issues/4504

## What does this do?

Makes the correspondence headers sticky so they're always visible as you scroll through the correspondence.

## Why was this needed?

So that when you're reading some correspondence, you can always know if it's incoming or outgoing, who it's from and when it was sent.

## Implementation notes

Wasn't sure if any bit of this would be in core rather than theme, but seemed to all fit nicely here anyway.

## Screenshots

(before I fixed rounded corners)

## Notes to reviewer

The rounded corners made it quite a bit more complicated because you don't want to see anything underneath the outside of the rounded corner. If you get rid of the rounded corners, most of it isn't needed.

### DIFF
--- a/app/assets/stylesheets/responsive/custom.scss
+++ b/app/assets/stylesheets/responsive/custom.scss
@@ -829,17 +829,17 @@ div.correspondence {
   box-shadow: 0 1px 2px rgba(0,0,0,0.25);
   border-radius: 5px;
   margin-bottom: 2.5em;
+  position: relative; // So we can position the rounded corners below
 
   &.collapsed {
     .correspondence__header {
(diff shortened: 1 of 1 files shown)


garethrees (approved):
So good! 🪄 

I think we'd want to port this in to core, but happy to merge here to see how we find it for a while.

lucascumsille (approved):

lucascumsille:
@dracos looks really great =) 
Tested it on Safari, Chrome and Firefox.

PR #1751, opened 2023-07-14
[#1750] Fix minor glitches in `about_foisa.html.erb`

### LINKED ISSUE #1750: [#1725] Fix minor glitches in `about_foisa`
Looks like some errors snuck in here - probably my fault.

1. The 'what is the…' header wraps when it ought not to. This seems to be because the anchor is on a new line, and is not spaced with `&nbsp;`
2. The last line states "on the OSIC on their website" - it should be one or the other.

mdeuk:
## Relevant issue(s)
Fixes #1750
#1725 

## What does this do?

1. Fixes the spacing in the first `h2` heading - to stop the anchor flowing onto a new line
2. Removes some duplicate wording in the copy, which could have caused confusion

## Why was this needed?

A couple of minor glitches snuck in when #1725 was pushed. Mea culpa.

## Implementation notes

Nothing of note here.

## Screenshots

| Before | After |
|:-------|:------------|
| Screenshot showing text in heading is split across two lines | Screenshot showing that the heading is now correctly displayed on a single line |

## Notes to reviewer

N/A

### DIFF
--- a/lib/views/help/about_foisa.html.erb
+++ b/lib/views/help/about_foisa.html.erb
@@ -156,7 +156,7 @@
   <p>
   You can also contact them for advice, by calling their helpline on <b>01334 464 610</b>, or if you use <abbr title="British Sign Language">BSL</abbr>, by using <a href="http://contactscotland-bsl.org">Contact Scotland BSL</a>.
   </p>
-  <p>There is more information on how to appeal, including a form that you can use (if you'd like), on the <abbr title="Office of the Scottish Information Commissioner">OSIC</abbr> on their website, at: <a href="https://www.itspublicknowledge.info/appeal">https://www.itspublicknowledge.info/appeal</a>.</p>
+  <p>There is more information on how to appeal, including a form that you can use (if you'd like), on their website, at: <a href="https://www.itspublicknowledge.info/appeal">https://www.itspublicknowledge.info/appeal</a>.</p>
   </p>
   <%= render partial: 'history' %>
   <div id="hash_link_padding"></div>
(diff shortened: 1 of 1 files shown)


gbp (approved):

lib/views/help/about_foisa.html.erb
```diff
@@ -156,7 +156,7 @@
   <p>
   You can also contact them for advice, by calling their helpline on <b>01334 464 610</b>, or if you use <abbr title="British Sign Language">BSL</abbr>, by using <a href="http://contactscotland-bsl.org">Contact Scotland BSL</a>.
   </p>
-  <p>There is more information on how to appeal, including a form that you can use (if you'd like), on the <abbr title="Office of the Scottish Information Commissioner">OSIC</abbr> on their website, at: <a href="https://www.itspublicknowledge.info/appeal">https://www.itspublicknowledge.info/appeal</a>.</p>
+  <p>There is more information on how to appeal, including a form that you can use (if you'd like), on their website, at: <a href="https://www.itspublicknowledge.info/appeal">https://www.itspublicknowledge.info/appeal</a>.</p>
```
garethrees, line 159: Only thing I'd say here is that we lose the one place we use the OSIC acronym (+ its definition), and I think it's nice to have that defined somewhere. Perhaps we could rephrase as:

> There is more information on how to appeal, including a form that you can use (if you'd like), on the OSIC's website at: https://www.itspublicknowledge.info/appeal.

garethrees (commented):
LGTM. Happy to merge as is, but one suggestion to consider.

lib/views/help/about_foisa.html.erb
mdeuk: That was intentional, because we didn’t use it elsewhere - but I see what you mean!

I think the best place to introduce it would be a little further up, and then we can use `abbr` tags elsewhere. I have a vague recollection of that being the initial intent… i think. 😳

garethrees: Nice!

garethrees:
Merged as dbaceee

Issue #1710, opened 2023-06-16
Add "Why FOI is useful" homepage section


garethrees:
Our front page says _what_ wdtk/foi is, and _how_ to use it (the 1,2,3 and some examples of bodies and recent requests), but not _why_.

It would be good to put together a short list of why FOI/WDTK is a great utility. We do this for the [pro use cases](https://www.whatdotheyknow.com/pro/pages/journalists) but we don't make a good case for the situations in which FOI might be a good tool more generally.

A couple that come to mind:

* It can reveal key facts and figures https://www.mysociety.org/2023/06/15/foi-helps-campaigns-sometimes-in-unexpected-ways/
* It can be used to compare and contrast https://www.mysociety.org/2023/06/07/freedom-of-information-uncovers-inequalities-in-our-schooling-system/
* It can show demand for proactive publication
* You can build a new dataset
* It can show how or why decisions got made (e.g. meeting minutes)
* It sews seeds for others to dig in to (i.e. they see something interesting in your request that you didn't explore further)
* Requesting in public builds a set of templates for others https://www.mysociety.org/2023/06/13/paul-bradshaw-and-whatdotheyknow/

We'll need to hone this list down to our key messages. 6-8 items.

In terms of display, I was thinking emoji + bold key point + very short explanation + maybe a link to relevant content, if we have it.

<img width="781" alt="Screenshot 2023-06-16 at 11 23 31" src="https://github.com/mysociety/whatdotheyknow-theme/assets/282788/1e86a550-e92e-41ff-9afe-d7f45e3731e0">

garethrees:
NB, we should probably sharpen the phrasing of "Why FOI is useful" – treat that as a working title.

The problem we're trying to tackle here is the kind of circular conversation like "FOI is great? Oh yeah, why? Because you can ask for anything and they more-or-less have to give it you you. Cool, but what should I ask for? Anything!"

Giving people mental models of approaches to take ("Find key facts and figures", "Compare and contrast", "Build a dataset", "Understand decisions", and so on) might help them apply FOI to their area of interest.

Hypothetically, let's say I'm interested in why my area of the city feels less green than others. Without a good understanding of these things, the best request I might be able to make is "Why does my area feel so concrete?".

With the above mental models, I might be able to construct a much better set of questions:

* Number of trees in my area (key fact)
* Number of trees in the fancy bit (compare & contrast)
* Number of tree protection orders in each (surface a dataset)
[...]

PR #1766, opened 2023-08-07
[#1710] FOI is awesome

garethrees:
## Relevant issue(s)

Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/1710

## What does this do?

Adds some homepage sections around:

* What is FOI
* Why is FOI useful
* How does WDTK help

## Why was this needed?

We don't really explain how a newcomer to FOI can use it.

## Implementation notes

Stole the CSS from the "articles" section. I think we could make more generic components here, but it'll do for now.

## Screenshots

foi-is-awesome

### DIFF
--- a/lib/views/general/_what_is_foi.html.erb
+++ b/lib/views/general/_what_is_foi.html.erb
@@ -1,0 +1,42 @@
+<div class="homepage-article-section" id="homepage-article">
+  <div class="row homepage-article-section__ar-contet">
+    <div class="ar-content__headings ">
+      <h2><%= _('What is Freedom of Information?') %></h2>
+
+      <p class="ar-content__lede">
+        <%= _('Make a request for information to a UK public authority.') %><br>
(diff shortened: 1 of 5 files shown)


lib/views/general/_what_is_foi.html.erb
```diff
@@ -0,0 +1,43 @@
...
+    <div class="ar-content__article-grid col-2-grid">
+      <div class="article-grid__item">
+        <div class="article-grid__item-content">
+          <p>
+            <%= _('The Freedom of Information Act, also known as <strong>FOI gives you the right to <a href="{{help_link}}">request recorded information</a> from public authorities</strong>.', help_link: help_requesting_path) %>
+          </p>
+
+          <p>
+            As well as documents and emails, it also covers things like
+            spreadsheets, TODO, TODO, videos and photographs.
```
garethrees, line 21: presentations,
shapefiles,
audio,
maps

Issue #1767, opened 2023-08-09
Data breach form does not set `reply-to` header


mdeuk:
When reports are received using the new data breach form, we are not seeing a `Reply-To` header, meaning the reply address has to be deciphered from the message body itself (if present).

This isn't a bug - but it would be helpful to look at, since we would like to be able to handle case creation via a partially automated workflow.

## Related
- https://github.com/mysociety/whatdotheyknow-theme/issues/1602
- https://github.com/mysociety/whatdotheyknow-theme/pull/1696

PR #1768, opened 2023-08-09
Set Reply-To for data breach emails.

dracos:
Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/1767

### DIFF
--- a/lib/data_breach.rb
+++ b/lib/data_breach.rb
@@ -53,6 +53,9 @@ module DataBreach
       from = MailHandler.address_from_name_and_email(
         'WhatDoTheyKnow.com data breach report', blackhole_email
       )
+      if report.contact_email
+        set_reply_to_headers(nil, 'Reply-To' => report.contact_email)
+      end
 
       # Set a header so we can filter in the mailbox
       headers['X-WDTK-Contact'] = 'wdtk-data-breach-report'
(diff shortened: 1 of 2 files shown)


lib/data_breach.rb
```diff
@@ -53,6 +53,7 @@ def data_breach(report, logged_in_user)
       from = MailHandler.address_from_name_and_email(
         'WhatDoTheyKnow.com data breach report', blackhole_email
       )
+      set_reply_to_headers(nil, 'Reply-To' => report.contact_email)
```
chrismytton, line 56: The `contact_email` field is optional, so we won't necessarily have an email address in `report.contact_email`.

We might also have a `logged_in_user`, so perhaps want to pass in the `logged_in_user` as the first argument, and then if there's an email address in `contact_email`, pass that in as the reply-to header?

Not sure exactly how the blackhole stuff works inside `set_reply_to_headers`, @gbp may be able to give more detailed guidance 🙂

chrismytton (changes requested):

lib/data_breach.rb
chrismytton: Yes on reflection I think you're right, best to leave the logged in user out of this.

chrismytton (approved):

garethrees:
Thanks both!

PR #1795, opened 2023-10-11
Update change request emails copy

gbp:
## Relevant issue(s)

Replaces https://github.com/mysociety/alaveteli/pull/7936

## What does this do?

Minor wording changes for new authorities/authorities change emails

## Why was this needed?

This change will reduce admin time editing the default outgoing message when accepting a request to add a new authority.


confirmordeny:
thank you

PR #1798, opened 2023-10-19
Set custom mySociety Datasette URL

garethrees:
Requires https://github.com/mysociety/alaveteli/pull/7961.

Screenshot 2023-10-19 at 17 56 08

### DIFF
--- a/lib/helper_patches.rb
+++ b/lib/helper_patches.rb
@@ -17,4 +17,6 @@ Rails.configuration.to_prepare do
       uri.to_s
     end
   end
+
+  DatasetteHelper.datasette_url = 'https://data.mysociety.org/datasette/'
 end


lib/helper_patches.rb
```diff
@@ -17,4 +17,6 @@ def request_url_with_query
       uri.to_s
     end
   end
+
+  DatasetteHelper.datasette_url = 'https://data.mysociety.org/datasette/'
```
gbp, line 21: I'm okay with this but would why like this instead of a `config/general.yml` configuration variable which we would normally use?

gbp (approved):
I'm okay with this but would why like this instead of a config/general.yml configuration variable which we would normally use?

garethrees:
> I'm okay with this but would why like this instead of a config/general.yml configuration variable which we would normally use?

Added some notes in https://github.com/mysociety/alaveteli/pull/7961/commits/a02cb120f1f8f8a20417173cbaedc461bb961a14, but to expand on that, there's also a complexity cost for adding to `general.yml`. It already has 115 keys, which is a bit 😵‍💫. I expect we'll be the only people to use this config, but if there ends up being demand we can always upgrade it from a theme initialisation config to `general.yml`.

PR #1814, opened 2023-12-15
[#1813] Breach form submitter contact email

### LINKED ISSUE #1813: Collect reporter's details on data breach form
Following #1777, I am now pretty confident that collecting the reporter's details on the data breach form is the right thing to do, so this should be considered a development ticket.

Therefore, contact details should be collected as on the other forms. This means using the details of the logged-in user, otherwise providing a name field and an email field to be filled in.

I think the `Reply-To` should be set to the form-filler, and not the DPO email address.

The DPO email field should remain separate and optional.

garethrees:
## Relevant issue(s)

Fixes #1813 
Depends on https://github.com/mysociety/alaveteli/pull/8074

## What does this do?

* Collect breach form submitter email

## Screenshots

Signed out:

Screenshot 2023-12-15 at 08 57 26

Signed in:

Screenshot 2023-12-15 at 08 57 12

### DIFF
--- a/lib/data_breach.rb
+++ b/lib/data_breach.rb
@@ -6,6 +6,7 @@ module DataBreach
     attribute :url, :string
     attribute :special_category_or_criminal_offence_data, :boolean
     attribute :message, :string
+    attribute :contact_email, :string
     attribute :dpo_contact_email, :string
     attribute :is_public_body, :boolean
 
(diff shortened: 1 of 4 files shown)


garethrees:
Think this is mostly there now. Only thing it could _ideally_ do with are some specs to cover the `@logged_in_user` case for the email body & Reply-To header.

gbp (changes requested):
Submitting the form while logged in doesn't work. It returns the error "Please include your email address".

gbp (approved):
@garethrees this should be good to merge now.

PR #1816, opened 2023-12-18
Send targeted request

confirmordeny:
## Relevant issue(s)
Send targeted requests is better than "Don't carpet bomb" for two reasons: (1) it is more accessible language and (2) it is likely to evoke images of violence.

## What does this do?
See above.

## Why was this needed?
See above.

## Implementation notes
See above.

## Screenshots
N/A.

## Notes to reviewer
See above.

### DIFF
--- a/lib/views/help/requesting.html.erb
+++ b/lib/views/help/requesting.html.erb
@@ -201,7 +201,7 @@
         </li>
         <li>
           <p>
-            <strong>Send targetted requests:</strong>
+            <strong>Send targeted requests:</strong>
             You might be planning to make a request to, for
             example, all local councils, or even all public bodies in the country,
             but  it’s worth stopping to reflect. Such bulk requests result in a
--- a/lib/views/help/requesting.html.erb
+++ b/lib/views/help/requesting.html.erb
@@ -201,7 +201,7 @@
         </li>
         <li>
           <p>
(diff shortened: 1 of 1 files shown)


lib/views/help/requesting.html.erb
```diff
@@ -201,7 +201,7 @@
         </li>
         <li>
           <p>
-            <strong>Don’t carpet-bomb:</strong>
+            <strong>Send targetted requests:</strong>
```
garethrees, line 204: typo:

```diff
- targetted
+ targeted
```

+1 on changing this; I've pointed to it a few times recently and grimaced a little when reading.

I wonder if this is a bit close to "Word your question carefully" and "Fine-tune your request".

This point is specifically trying to avoid bulk requesting in a way that's disruptive – the word scattergun comes to mind, but annoyingly similar connotations. Might be a nice alternative to [indiscriminate](https://www.wordhippo.com/what-is/another-word-for/indiscriminate.html)? If we can't think of anything else then happy with "Send targeted requests".

garethrees (changes requested):

lib/views/help/requesting.html.erb
confirmordeny: thank you

PR #1818, opened 2024-01-09
[WDTK] Updated mySociety pennant

lucascumsille:
## Relevant issue(s)
https://github.com/mysociety/Design-Internal-tasks/issues/35

## What does this do?
Replaces 20th anniversary pennant with a normal mySociety one.

## Screenshots
Screenshot 2024-01-09 at 13 58 50

Screenshot 2024-01-09 at 13 58 36


garethrees:
Screenshots look good – quick question; the dimensions of the new image isn't the same. Assuming this is intentional and okay?

lucascumsille:
Thanks @garethrees actually they are smaller, but I just pushed a commit to actually optimise the size as well. The ratio should be the same though.

garethrees (approved):

PR #1837, opened 2024-02-28
Patch authority only response gatekeeper

gbp:
## What does this do?

Patch authority only response gatekeeper

## Why was this needed?

Accept responses to authority only requests from extra addresses which authorities might use. For example no-reply addresses from case management systems which aren't changeable.

## Implementation notes

Made it so adding other addresses is trivial so volunteers can open PRs for other addresses like they do for `invalid_reply_addresses`


gbp:
@garethrees this seems to work well. Can you see any issues which I might be missing with this approach?

PR #1845, opened 2024-03-25
Add Bournemouth, Christchurch and Poole Council no reply address

garethrees:
Adding due to known issue of escaping ampersands:

https://github.com/mysociety/alaveteli/issues/3465


gbp:
Deployed

PR #1854, opened 2024-04-19
Topic pages

gbp:
## Relevant issue(s)

https://github.com/mysociety/alaveteli/issues/7981

## What does this do?

1. Add subnav styling
2. Add feature flag for topic pages

## Why was this needed?

Needed for https://github.com/mysociety/alaveteli/pull/8206

## Screenshots

Screenshot 2024-04-19 at 13-03-09 Search requests
Screenshot 2024-04-19 at 13-03-01 Browse requests by category


garethrees:
I think we could add this to core so that we/re-users don't have to add it in for each theme.

PR #1873, opened 2024-05-16
Exclude schools from batch requests

gbp:
## Relevant issue(s)

Depends on: https://github.com/mysociety/alaveteli/pull/8258

## What does this do?

Exclude schools from batch requests

## Why was this needed?

This previously was done in Alaveteli core for the batch category UI. This will now also exclude schools from the batch authority search.


garethrees:
Sounds good until we resolve https://github.com/mysociety/whatdotheyknow-theme/issues/1874.

PR #1884, opened 2024-05-31
[ExcelAnalyzer] republish attachments automatically

gbp:
If the attachment doesn't get flagged as containing potential PII there are cases where we're comfortable in republishing the attachment automatically, exceptions are:
1. If XLS with pivot cache - the conversion to XLSX means the pivot cache is lost
2. With named ranges and external links present - the data cached isn't being scanned for PII
3. With a data model present - we can't, at present, scan the data inside this data model.

Fixes https://github.com/mysociety/whatdotheyknow-private/issues/312

### DIFF
--- a/lib/excel_analyzer/jobs/republish_or_report_job.rb
+++ b/lib/excel_analyzer/jobs/republish_or_report_job.rb
@@ -1,0 +1,80 @@
+##
+# Job to republish or report to admins files stored as ActiveStorage::Blob
+#
+# Examples:
+#   ExcelAnalyzer::RepublishOrReportJob.perform(ActiveStorage::Blob.first)
+#
+module ExcelAnalyzer
+  class RepublishOrReportJob < ApplicationJob
+    queue_as :excel_analyzer
+
+    attr_reader :attachment_blob
+
+    def perform(attachment_blob)
+      @attachment_blob = attachment_blob
+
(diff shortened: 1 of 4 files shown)


lib/excel_analyzer/jobs/republish_or_report_job.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    attr_reader :attachment_blob
+
+    def perform(attachment_blob)
+      @attachment_blob = attachment_blob
+
+      if !potential_pii_leaks? &&
+        !(xls? && pivot_cache?) &&
+        !(named_ranges? && external_links?) &&
+        !(data_model?)
```
garethrees, line 19: Would make this

```rb
if safe_to_republish?
  attachment.update_and_log_event(
  #…

def safe_to_republish?
  !(xls? && pivot_cache?) &&
    !(named_ranges? && external_links?) &&
    !(data_model?)
end
```

gbp: Just misisng the `potential_pii_leaks` check be yes could extract out to a `safe_to_republish?` helper

garethrees: oops yeah but the main thing was making it a bit clearer what the big condition means

PR #1895, opened 2024-07-05
WIP: Add citation promo to pro mailouts

garethrees:
Fixes https://github.com/mysociety/whatdotheyknow-private/issues/332.

Adding only the HTML part to the theme seems to prevent the mail being sent as multipart. IDK whether this is an issue with the rails previewer not working well with our theme override, or whether this will also happen in production with it stopping at the theme `embargo_mailer` directory instead of falling back to the core one to get the text version. If so, we'll just have to override the text template without modification.

Obviously need to add the link to the blog post, and maybe tweak phrasing / design of the banner.

Screenshot 2024-07-05 at 16 17 16

Screenshot 2024-07-05 at 16 17 29


garethrees:
Did look at using a layout for the promo banner but that would mean customising all methods to set the layout in the format block; doesn't seem like its possible/easy to set the layout for any html part at the class level of the mailer.

PR #1896, opened 2024-07-15
Fix button-tertiary styling

gbp:
Ensure these button are the same height as primary buttons.

Before (with button hover to show difference):
image

After:
image

The remaining difference is sue to the primary button boarder styling:
image


garethrees (approved):
🙌

PR #1902, opened 2024-07-31
[ExcelAnalyzer] Patch RequestMailer.new_response

gbp:
## Relevant issue(s)

Fixes https://github.com/mysociety/whatdotheyknow-private/issues/315

## What does this do?

Patch RequestMailer.new_response

## Why was this needed?

Don't send no response notifications while we're processing attachments, instead wait until the ExcelAnalyzer and PIIBadger scripts have run to determine if the attachments in the response have been hidden on not.


gbp:
mail-receive-flow

PR #1917, opened 2024-09-09
[WDTK] Improved focus state for buttons

lucascumsille:
## Relevant issue(s)
Discovered while working on a blog about accessibility for WDTK

## What does this do?
Makes the focus state more obvious for users with visual impairments

## Screenshots

https://github.com/user-attachments/assets/d749e787-c6c2-4923-8b26-08342414f27e

### DIFF
--- a/app/assets/stylesheets/responsive/_mixins.scss
+++ b/app/assets/stylesheets/responsive/_mixins.scss
@@ -7,13 +7,26 @@
 
 
 /* Buttons */
+@mixin focus-state {
+  &:focus,
+  &:visited:focus {
+    // It overrides any specificity like the Pro buttons
+    background: $color_yellow !important;
+    color: $color_black !important;
+
+    outline: 3px solid $color_off_white; // Most of the time this won't be visible, but it will be helpful when the button is in a dark background.
+    outline-offset: 0;
+    box-shadow:  inset 0px 0px 0px 2px $color_black;
+  }
+}
+
 @mixin button-base {
   padding: 0.5em 1.25em;
   border-radius: 3px;
   display: inline-block;
   font-weight: 600;
   text-decoration: none;
-  transition: background-color 300ms ease-out;
+  transition: background 300ms ease-out;
   line-height: normal;
   margin-bottom: 0;
   font-family: $font-family;
@@ -21,8 +34,10 @@
   &:active,
   &:focus {
       text-decoration: none;
-      transition: background-color 300ms ease-out;
+      transition: background 300ms ease-out;
   }
+
+  @include focus-state;
 
   &:disabled {
     @include button-disabled;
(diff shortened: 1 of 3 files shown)


garethrees:
So nice! ✨

gbp (changes requested):
There are a few buttons I've found which don't have the new focus state.

All Pro styled buttons:
Screenshot 2024-09-16 at 09 42 23

Donate button in the footer:
Screenshot 2024-09-16 at 09 42 54

And the homepage "Make a request" button unfocused state is yellow so there isn't much difference when focused:
Screenshot 2024-09-16 at 09 44 15

gbp (commented):
@lucascumsille can I check if the focus state should be shown on all footer anchors or not. Seems a bit strange...


https://github.com/user-attachments/assets/1b130957-d3c8-4354-a4b7-60f415607c3a

gbp (commented):
@lucascumsille also the yellow button on the homepage... looks like you've increased the size of boarder but haven't changed the colour of the button. Are you happy with the focus state on that now?

lucascumsille:
@gbp I have included the changes made on [this PR](https://github.com/mysociety/whatdotheyknow-theme/pull/1918) on the current PR instead. I think with these commits, we should be covering all cases.

As a summary:

- Added transition effect to footer button.
- Added a more visible focus state. Using a light outline and a dark `box-shadow` with no blur, we cover scenarios where the button background has a different level of lightness.
The default button has a hover effect and a background that is slightly different from white, so when hovering over it right next to a tertiary button, the user doesn't get confused.
Remove transitions from hover effects. We don't need to duplicate them; as long as the transition is in the default state, the other states will inherit it.
The styling for hover and focus has changed so that it changes only two properties compared with the default button, making it clearer for users with visual impairments.
- The hover effect for the links in the footer wasn't accessible, so I made them lighter.


**Pending(Potentially):**

>can I check if the focus state should be shown on all footer anchors or not. Seems a bit strange...
- At the beginning, I included the focus state for links as well; I took as an inspiration [GOVUK](https://design-system.service.gov.uk/styles/links/). It's not so strange, and might not be the most aesthetically pleasing effect, but it's familiar for users that use the keyboard to navigate. I didn't include it in this PR, though, because in your video, I noticed that it messed up links with logos(footer and navbar). But if you think those are the only two cases where we might end up with a problem, then it might be easier to add a rule to exclude those classes with logos(assuming we are happy to move ahead with the effect for the focus state with yellow bg).

We are currently using transitions for buttons. Do you want me to include one for the links as well? I was wondering if there was a reason why it wasn't included initially.


### Preview


https://github.com/user-attachments/assets/f6bc88af-ceda-4c3c-a2fc-bf5aa5715cc1


https://github.com/user-attachments/assets/4ecd7162-3a3b-4921-9c53-9f0833a9bac2


https://github.com/user-attachments/assets/c0b122f5-51d4-4819-9b93-38ab79639fd8


https://github.com/user-attachments/assets/b0e872d9-e215-4e95-a05a-d3da177cdf87

gbp (commented):
This is looking good. Only one point, the preview video of the yellow homepage button doesn't look like what I'm seeing. It is first going to dark blue then suddenly to the lighter shade of blue. Where there any changes made after the video was made or could it be an issue with the browser I'm using?

lucascumsille:
@gbp the :visited was overriding the styling, I added a fixup that should prevent that from happening.

https://github.com/user-attachments/assets/a5a66c06-aab2-4b62-aecf-48ce4784a45f

gbp (approved):
Amazing - thank you 👍

lucascumsille:
@gbp Thanks =) I just rebased, but would you prefer me to make everything one commit or maybe two(buttons and one for the footer)?

gbp:
> @gbp Thanks =) I just rebased, but would you prefer me to make everything one commit or maybe two(buttons and one for the footer)?

Whatever you feel makes the best sense, even if it's just the 4 commits as they are.

PR #1918, opened 2024-09-10
[WDTK] Improved visual elements for button classes

lucascumsille:
## Relevant issue(s)
No issue, the default state of the tertiary button gave me the idea to make some improvements to some of the default and for all the hover states.

## What does this do?
Currently the hover effect is quite subtle, which is not ideal for users with visual impairments. Ideally we want to change at least two properties when changing between states, to make it fairly obvious. For the tertiary button, I also changed the default state, to make more evident that is an interactive element. For this I increased the border-width, and changed the color(border and font) to what we use for links, which makes it more familiar for users that is an interactive element.

## Screenshots

**Default**

https://github.com/user-attachments/assets/841643d7-b3ca-4366-a892-30b187dd3c73

**Pro**

https://github.com/user-attachments/assets/96e406a6-ea87-41e1-b486-348ff36c01fe

### DIFF
--- a/app/assets/stylesheets/responsive/_mixins.scss
+++ b/app/assets/stylesheets/responsive/_mixins.scss
@@ -12,6 +12,7 @@
   &:visited:focus {
     background-color: $color_yellow;
     color: $color_black;
+    background: $color_yellow !important;
   }
 }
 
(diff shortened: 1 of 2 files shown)


gbp:
@lucascumsille we also have a style of buttons for the Pro interface. Do we need an update there too:

image

gbp (changes requested):
The hover transition for the primary button seems to jump to grey. It isn't as smooth as the other button types. Is it possible to fix this?

https://github.com/user-attachments/assets/e64eb547-68b8-4e30-b1ae-d1768dcd0faf

lucascumsille:
@gbp I added a mixup that should solve the transition issue. Btw I erase some of the transitions that have been duplicated for hover effects, the bit transition is only needed it in the default and then it will get use on all the other states.

gbp (commented):
@lucascumsille this have merge conflicts and doesn't merge cleanly. I believe I've merged it locally by manually fixing the conflicts but need to check with you because some hover states which look strange:

https://github.com/user-attachments/assets/768d4b01-221b-4534-b2ed-3949b3b68e58


https://github.com/user-attachments/assets/8058a2aa-1cb7-4066-bd59-59718c08c439


https://github.com/user-attachments/assets/4c3dcf36-7e27-485f-9f56-2511086e2673

lucascumsille:
@gbp No worries I'll do the merging, I think we are losing some stuff there.

lucascumsille:
Replaced by https://github.com/mysociety/whatdotheyknow-theme/pull/1917

Issue #1920, opened 2024-09-20
Banner colour pairing fails accessibility checks


HelenWDTK:
The white on green text is hard for me to read, and fails contrast tests.

<img width="627" alt="Screenshot 2024-09-20 at 09 54 40" src="https://github.com/user-attachments/assets/4fc117c2-b5fb-42db-92ca-55970621c2a7">

PR #1923, opened 2024-09-23
[WDTK] Popup black version

lucascumsille:
## Relevant issue(s)
[Helen discovered ](https://github.com/mysociety/whatdotheyknow-theme/issues/1920)

## What does this do?
Improved contrast for the popup element so it passes contrast text

## Screenshots

**Black version**
Screenshot 2024-09-23 at 07 44 07

**Green version**
I get why originally we are using the mySociety green.  but I think the problem with that it rarely looks nice inmediately right next to the other colours, they tend to clash with each other. Assuming using green is important I added a gradient that makes it readable, but probably my least favourite option. I tried a darker version so we can continue using white font text, but it didn't look very nice next to the navbar.
Screenshot 2024-09-23 at 07 59 32

**WDTK Variation**
I created a variation of WDTK primary colour, using the same HUE, unlike the green version I think this one looks more WDTK style. 
Screenshot 2024-09-23 at 08 14 20

Let me know if you have any preferences.

### DIFF
--- a/app/assets/stylesheets/responsive/custom.scss
+++ b/app/assets/stylesheets/responsive/custom.scss
@@ -317,14 +317,14 @@ a.link_button_green_large {
 /* Popups */
 
 .popup {
-  background-color: $color_black;
+  background: linear-gradient(90deg, rgba(140, 210, 130, 1) 0%, rgba(192, 246, 185, 1) 100%);
   border: 0;
-  color: $color_white;
+  color: $color_black;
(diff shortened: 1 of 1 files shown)


HelenWDTK:
I love the WDTK Variation - I don't think there is any particular reason why it needs to be green.

lucascumsille:
Thanks you @HelenWDTK happy to go with that. The first one(Black version) is a bit subtle, while the WDTK versions pop up more.

PR #1933, opened 2024-11-18
Update model_patches.rb

confirmordeny:
## Relevant issue(s)
None 

## What does this do?
Added gmail, googlemail and outlook to list so they are not used to calculate homepage

## Why was this needed?
More complete list

## Implementation notes
N/A

## Screenshots
N/A

## Notes to reviewer
N/A

### DIFF
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -84,9 +84,12 @@ Rails.configuration.to_prepare do
     btconnect.com
     btinternet.com
     btopenworld.com
+    gmail.com
+    googlemail.com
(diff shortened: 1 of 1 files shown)


gbp (commented):
This is not needed. These domains are already [included in core Alaveteli](https://github.com/mysociety/alaveteli/blob/39d7998653f41c9addaf79044478bff8a4345e28/app/models/public_body/calculated_home_page.rb#L6-L25). The ones listed here are UK specific domains which are appended to the core list.

Issue #1907, opened 2024-08-12
Add line to House Rules to clarify that WDTK is not a general publishing platform


HelenWDTK:
Something along the lines of WDTK is only for making requests for information and not for publishing your own views, but better.

confirmordeny:
... and not a mechanism for sending information to public authorities or publishing information you already hold.

WilliamWDTK:
The trouble with saying it's "only for asking public authorities for information you don't have" is that someone may legitimately wish to bring information to light that they have insider/unofficial knowledge of, which would technically fall foul of not asking for information you already have.

HelenWDTK:
I don't think knowledge of is the same as already have, but then I am very literal, so I take the point. 

We just need to make it clear that it's not for reposting things that people have already got or publishing in general.

I've got a window of time so if we can come up with a form of words by the end of week, I can get this done, if not, it'll likely be next year.

garethrees:
I think "Only use WhatDoTheyKnow to request specific information. Don't use it to send general messages to public authorities or for personal messages." does cover this, but we could add some examples.

Could try something like:

"FOI requests are for requesting facts, figures, data, and recorded information. A _small_ amount of additional context should only be included if it is necessary to help the authority identify the information. Avoid including opinions, complaints, or data about yourself in your requests."

PR #1936, opened 2024-11-23
Update house_rules.html.erb

confirmordeny:
## Relevant issue(s)
Issue #1907

## What does this do?

I have proposed some changes to the House Rules.

## Why was this needed?

Hopefully, this clarifies how users should use the site and reduces misuse of the site.


confirmordeny:
This relates to issue #1907

Issue #1741, opened 2023-07-10
ATI books help page


garethrees:
**See list at https://github.com/mysociety/whatdotheyknow-theme/issues/1703**. Why do this? Help grow the community of practice and signal-boost those who've put the time and effort into documenting their knowledge.

This could start life as a help page under the new [access to information help](https://github.com/mysociety/whatdotheyknow-theme/pull/1738#issuecomment-1622139985) section.

Doesn't need to be anything flashy. Just book title, cover image, a few lines of the blurb and a link to where someone can get it.

![Screenshot 2023-07-10 at 11 16 24](https://github.com/mysociety/whatdotheyknow-theme/assets/282788/64378efa-d783-48f5-b88d-196569d7720c)

garethrees:
> Do we have a list of titles in mind?

We do! 📚

> See list at https://github.com/mysociety/whatdotheyknow-theme/issues/1703.

HelenWDTK:
This issue is being closed due to a lack of discussion or resolution for over 12 months. Should we decide to revisit this issue in the future, it can be reopened.

PR #1943, opened 2024-12-18
Add /help/books

garethrees:
Add a list of recommended ATI books.

Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/1741.

Access to Information Books -  DEV  WhatDoTheyKnow ()


mdeuk:
> This seems like a valuable addition. Further to the [discussion on Slack](https://mysociety.slack.com/archives/CNH83HJP4/p1734539061476639), I think it could be worthwhile:
> 
>     1. Linking to the books on WorldCat, to encourage borrowing for those that might be interested, or that can't afford the more expensive books.

+1 - I would, perhaps, also go a little further to include a line about checking whether or not a book might be available at a public  library - either in print, or as an eBook.

That could be a useful option for those who don't have easy access to an academic library.

lib/views/help/books.html.erb
```diff
@@ -0,0 +1,87 @@
...
+
+<%= render partial: 'sidebar' %>
+
+<div id="left_column_flip" class="left_column_flip">
+
+  <div id="help_books">
+    <div class="box-out">
+      <div class="box-in">
+        <h1>Access to Information Books</h1>
+        <p>Our list of recommended books about Freedom of Information and how to use it.</p>
```
mdeuk, line 11:
```suggestion
        <p>Our list of recommended books about Freedom of Information and how to use it.</p>
        <p>
          Many of these books will be available, for free, in public libraries.
          <br>
          Not sure where your nearest library is? You can use the 
          <a href="https://libraryon.org/libraries">LibraryOn</a> website to find services close to you.
        </p>
```

garethrees: Nice suggestion – updated with a version of this.

garethrees:
Thanks both! I've added a version of a more general pointer to libraries in the intro paragraph. Going to just go with this as otherwise this will never get merged 🙈

PR #1945, opened 2024-12-19
Update model_patches.rb

### LINKED ISSUE #1944: Add Do Not Reply for department for education to the list
They are acknowledging requests from an unmonitored account

HelenWDTK:
Adds unmonitored DfE address to the list

## Relevant issue(s)
Fixes #1944 
## What does this do?
Adds the unmonitored address to the list
## Why was this needed?
If people write to it, the message won't be seen
## Implementation notes
n/a
## Screenshots
n/a
## Notes to reviewer
n/a

### DIFF
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -202,6 +202,7 @@ Rails.configuration.to_prepare do
     noreply@ams-sar.com
     foi@cheshireeast.gov.uk
     OneTrustEmail@surrey.ac.uk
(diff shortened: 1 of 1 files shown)


mdeuk (changes requested):
IIRC this is a rather misleading one, as replies to that mailbox have, historically, been processed, despite the _obvious_ name.

It might be worth us reaching out before merging this.

garethrees:
Is this one still needed?

gbp:
Closing but reopen if still needed.

PR #1947, opened 2024-12-27
Update _more_info.html.erb to support UCAS links

confirmordeny:
Add UCAS links on authority pages

## Relevant issue(s)
N/A

## What does this do?
Adds a link to UCAS website on each authority page for which there is an UCAS code.

## Why was this needed?
N/A

## Implementation notes
N/A

## Screenshots
N/A

## Notes to reviewer
Here is an example URL: https://www.ucas.com/explore/search/providers?query=A60

### DIFF
--- a/lib/views/public_body/_more_info.html.erb
+++ b/lib/views/public_body/_more_info.html.erb
@@ -74,6 +74,13 @@
   <% end %>
 <% end %>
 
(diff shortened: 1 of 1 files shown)


gbp (approved):
Might need to review out-of-date tags. For example we have an authority tagged `ucas:A14` but there is no result on the provider search.

Looks good otherwise.

PR #1956, opened 2025-01-10
Update _frontpage_videos.html.erb

### LINKED ISSUE #1955: Update list of videos on homepage
It's been a while since we've updated the list of videos imported into  the WDTK homepage and there's some great new content available.

There are some good transparency related TICTeC sessions, some specific international events, and the new video case study from the Our City book

For each video we need the title and the video ID:

https://github.com/mysociety/whatdotheyknow-theme/blob/master/lib/views/general/_frontpage_videos.html.erb#L77-L78

If someone could pull out those details for relevant videos that would be a big help for getting these updated.

WilliamWDTK:
Update videos from mySociety YouTube, as per #1955

## Relevant issue(s)
Fixes #1955 

## What does this do?
Adds some more relevant videos from mySociety's YouTube channel to the group that videos are selected from for the WDTK homepage.

## Why was this needed?
To maintain a more up-to-date homepage

### DIFF
--- a/lib/views/general/_frontpage_videos.html.erb
+++ b/lib/views/general/_frontpage_videos.html.erb
@@ -75,7 +75,25 @@
       { title: 'Creating datasets from FOI data',
         id: 'lar5OpiJsDk' },
       { title: 'Data Visualisation Show and Tell',
-        id: 'aKBoi-Qy9sE' }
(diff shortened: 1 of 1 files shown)


garethrees (approved):
Looks great to me in principle! Have tagged @gbp to double check the code and get it merged/deployed.

PR #1960, opened 2025-01-27
[#1840] Donation link

### LINKED ISSUE #1840: Make the link with mySociety more prominent on the mailouts
This is part of a wider push to ensure that users of each of our services know that it is one of many provided by mySociety.

Is it possible to add the 'a service from mySociety' pennant at the top of these mailings? I think they're plain text so might have to think of something else?

- Email alerts
- 'You have a response' emails
- Any other emails that we send in the course of a user transaction (sorry, did try looking at the code but I can't find them here or in Alaveteli!)

If we can't, perhaps we could look at how to reposition
>WhatDoTheyKnow is a project of mySociety run by a small team of staff and dedicated volunteers.

Which at the moment is just right at the bottom, in small text and easy to miss.

mdeuk:
## Relevant issue(s)

#1840

## What does this do?

This adds a new route, `/donate`, which points toward the WDTK specific donation page.

It also adds the enhanced "please donate" wording, from https://github.com/mysociety/whatdotheyknow-theme/issues/1840#issuecomment-2023092325

## Why was this needed?

Current mailbox signature has a rather cumbersome link, this enables a neater solution.

Additionally, the enhanced "ask" is hopefully more eye catching, within the limit that plaintext email has.

## Implementation notes

Nothing new here, it's just enhancing what is already in the code.

## Notes to reviewer

I'm not sure if this closes #1840 completely, or not, so haven't marked it as such.

### DIFF
--- a/lib/views/general/_default_footer.text.erb
+++ b/lib/views/general/_default_footer.text.erb
@@ -3,8 +3,8 @@
 WhatDoTheyKnow is a project of mySociety run by a small team of staff and
 dedicated volunteers.
 
-You can help to make information accessible to all by donating today.
-Your donations, however small, really help:
+mySociety is a charity that believes that information should be accessible to everyone.
+If this is important to you, please consider donating to help us do more:
 <%= donate_to_mysociety_url %>
 
 Find out more about how we handle your data at:
--- a/lib/views/general/_default_footer.text.erb
+++ b/lib/views/general/_default_footer.text.erb
@@ -3,11 +3,12 @@
 WhatDoTheyKnow is a project of mySociety run by a small team of staff and
 dedicated volunteers.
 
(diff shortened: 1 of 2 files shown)


lib/views/general/_default_footer.text.erb
```diff
@@ -3,8 +3,8 @@
 WhatDoTheyKnow is a project of mySociety run by a small team of staff and
 dedicated volunteers.
 
-You can help to make information accessible to all by donating today.
-Your donations, however small, really help:
+mySociety is a charity that believes that information should be accessible to everyone.
+If this is important to you, please consider donating to help us do more:
```
mdeuk, line 7: I realise this is a little long, but it didn't seem like it would flow as well if there was another linebreak (e.g. before 'accessible').

PR #1962, opened 2025-02-04
[core#8535] Pro rate limit upsell

gbp:
## Relevant issue(s)

Depends on https://github.com/mysociety/alaveteli/pull/8575

## What does this do?

Adds Pro rate limited upsell

## Screenshots

image

### DIFF
--- a/lib/views/user/rate_limited.html.erb
+++ b/lib/views/user/rate_limited.html.erb
@@ -1,50 +1,94 @@
-<% @title = _('Too many requests') %>
+<% @title = _("Too many requests") %>
 
-<h1><%= @title %></h1>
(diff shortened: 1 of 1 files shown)


garethrees (commented):
V1 lgtm.

I agree with @lucascumsille's comments. Would be good to better distinguish between the rate limited message and the upsell and use a CTA button for the "subscribe" link. The upsell doesn't need to be _really_ shouty though – it's incidental content for that page.

I'd move "Commercial and for-profit use requires a WhatDoTheyKnow Pro subscription" to be the last line. That way everything above can be put in core, and our override can just add that at the end.

PR #1973, opened 2025-03-11
Make some initial update to reference the Online Safety Act

sagepe:
This makes some initial additions to help pages with reference to the Online Safety Act. Further, more detailed, updates may be needed in due course.

### DIFF
--- a/lib/views/help/removing_information.html.erb
+++ b/lib/views/help/removing_information.html.erb
@@ -44,16 +44,42 @@
     harassment.</li>
     <li>Material that unfairly damages the reputation of a person, a business, or
     an organisation (sometimes called defamatory or libellous material).</li>
-    <li>Material that it is against the law to publish.</li>
+    <li>As defined in the Online Safety Act (2023), material that is terrorism
+    content, CSEA content and other priority illegal content.</li>
+    <li>Material more widely that it is against the law to publish.</li>
     <li>Any other material that would breach our House Rules.</li>
   </ul>
 
   <p>You can find a link to report a request on every request page.</p>
 
-  <h2 id="officers">
-    For FOI Officers and other employees of public authorities
-    <%= link_to '#', '#officers' %>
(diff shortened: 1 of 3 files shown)


garethrees (approved):
lgtm as a first pass. fixed a missed closing tag.

WilliamWDTK:
In one place, `CESA` is used, and in another `CSEA` is used. I assume these should both be the same.

lib/views/help/removing_information.html.erb
```diff
@@ -44,17 +44,43 @@
...
 
+  <h3 id="illegal">
+    I would like to report the presence of illegal content that should
+    be removed
+    <%= link_to '#', '#illegal' %>
+  </h3>
+
+  <p>We do not tolerate the publication of illegal material on the site, in particular
+  that which falls under the scope of the Online Safety Act (2023). This includes
+  terrorism content, CESA content and other forms of priority illegal content as defined
```
WilliamWDTK, line 68: I think `CESA` here should be `CSEA`.

mdeuk: Yeah, it [should be CSEA](https://www.ofcom.org.uk/online-safety/protecting-children/tackling-child-sexual-abuse-under-the-online-safety-regime/). It's probably a good idea if we spell the acronym out too.

```suggestion
  terrorism content, <abbr title="Child Sexual Exploitation and Abuse">CSEA</abbr> 
  content and other forms of priority illegal content as defined
```

WilliamWDTK: Would it be better to spell it out in full in the plain text, and just use an `<abbr>` tag on subsequent occurrences of the abbreviation?

mdeuk: I think that the former would generally be better - spell it out, then use the abbreviation. #1974 hopefully covers this off.

PR #1974, opened 2025-03-11
[#1973] Minor improvements to OSA content

mdeuk:
## Relevant issue(s)
#1973 

## What does this do?

This corrects a typo (`CESA` should be `CSEA`), and makes some slight changes to the copy, so that it flows a bit better. It also spells out `CSEA`, and references Ofcom with their common name.

## Why was this needed?

The copy was good, but it did not flow very well on user's screens, so it looked a bit like a wall of text (within one!), and there were a couple of typos (ref https://github.com/mysociety/whatdotheyknow-theme/pull/1973#pullrequestreview-2675954881).

Additionally, we referred to Ofcom as "Office of Communications" - we should try and avoid `govspeak` here, and use the more common name.

## Implementation notes

Nothing to note.

## Screenshots

| From | To |
|:------|:----|
| Screenshot showing the original rendering of the #illegal copy on page | Screenshot showing the proposed #illegal content, as shown in this PR |

## Notes to reviewer

The HTML on `removing_information.html.erb` is not consistently formatted. I've reformatted the bits in play here, but you might want to look at it more generally.

### DIFF
--- a/lib/views/help/complaints.html.erb
+++ b/lib/views/help/complaints.html.erb
@@ -46,7 +46,8 @@
       </p>
       <p>
         WhatDoTheyKnow’s activity is subject to regulation by the Information
-        Commissioner, the Office of Communications  and the Charity Commission.
+        Commissioner, <abbr title="Office of Communications">Ofcom</abbr>, and 
+        the Charity Commission.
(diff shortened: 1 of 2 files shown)


garethrees (approved):
Nice thanks!

PR #1983, opened 2025-04-05
Add link to ONS explore local statistics pages

### LINKED ISSUE #1982: Add link to ONS explore local statistics page for local council areas
These pages are really nice:

https://explore-local-statistics.beta.ons.gov.uk/areas/E09000004

The authorities are tagged up with the right codes already, so it would seem a shame not to link to these.

FOIMonkey:
## Relevant issue(s)
Fixes #1982 
## What does this do?
Adds a link to the ons local stats page
## Why was this needed?
We didn't link to is
## Implementation notes
n/a
## Screenshots
Screenshot 2025-04-05 at 15 42 14
## Notes to reviewer
I popped it above mentions in parliament, as it seemed more relevant, but can be relegated if that doesn't make sense.

### DIFF
--- a/lib/views/public_body/_more_info.html.erb
+++ b/lib/views/public_body/_more_info.html.erb
@@ -14,6 +14,13 @@
   <%= link_to _('Disclosure log'), public_body.disclosure_log %><br>
 <% end %>
 
+<% if public_body.has_tag?('statistical_geography') %>
+  <% public_body.get_tag_values('statistical_geography').each do |tag_value| %>
+      <%= link_to _('Explore local statistics'),
+                    "https://explore-local-statistics.beta.ons.gov.uk/areas/#{ tag_value }" %><br>
+  <% end %>
+<% end %>
+
 <% twfy_utm_params = { utm_source: 'whatdotheyknow.com',
                        utm_medium: 'link',
                        utm_campaign: 'body_page_more_info' } %>
--- a/lib/views/public_body/_more_info.html.erb
+++ b/lib/views/public_body/_more_info.html.erb
@@ -14,10 +14,18 @@
   <%= link_to _('Disclosure log'), public_body.disclosure_log %><br>
 <% end %>
 
+<% ons_statistics_tags = [] %>
 <% if public_body.has_tag?('statistical_geography') %>
-  <% public_body.get_tag_values('statistical_geography').each do |tag_value| %>
-      <%= link_to _('Explore local statistics'),
-                    "https://explore-local-statistics.beta.ons.gov.uk/areas/#{ tag_value }" %><br>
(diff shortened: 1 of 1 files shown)


FOIMonkey:
I think there is overlap with GSS codes, but am not 100% sure. I got the data from this dataset that seems to autoupdate (I know the gov was about to stop maintaining some of these, so that might not continue): https://www.planning.data.gov.uk/dataset/local-authority

lib/views/public_body/_more_info.html.erb
```diff
@@ -14,6 +14,13 @@
   <%= link_to _('Disclosure log'), public_body.disclosure_log %><br>
 <% end %>
 
+<% if public_body.has_tag?('statistical_geography') %>
+  <% public_body.get_tag_values('statistical_geography').each do |tag_value| %>
+      <%= link_to _('Explore local statistics'),
+                    "https://explore-local-statistics.beta.ons.gov.uk/areas/#{ tag_value }" %><br>
+  <% end %>
+<% end %>
```
mdeuk, line 22: Our :robot: overlords have suggested that we can use a bit of recursion to handle cases where either tag (or both exists… 

```suggestion
<% ons_statistics_tags = [] %>
<% if public_body.has_tag?('statistical_geography') %>
  <% ons_statistics_tags += public_body.get_tag_values('statistical_geography') %>
<% end %>
<% if public_body.has_tag?('gss') %>
  <% ons_statistics_tags += public_body.get_tag_values('gss') %>
<% end %>

<% unless ons_statistics_tags.empty? %>
  <% ons_statistics_tags.each do |tag_value| %>
    <%= link_to _('Explore local statistics'),
          "https://explore-local-statistics.beta.ons.gov.uk/areas/#{ tag_value }" %><br>
  <% end %>
<% end %>
```

mdeuk (commented):
I've tested this on a sandbox, and it works great.

I've added a suggested change, which the bot helped with - covering:
1. Cases where `statistical_geography`, and `gss` exist[^1]
2. Changing the URL from the beta to the production URL [^2]
3. Adds some [encoding logic](https://rubyapi.org/3.4/o/erb/util), [courtesy of the bot](https://github.com/mysociety/whatdotheyknow-theme/pull/1983#discussion_r2029892163).

… feel free to ignore, or re-implement, as you see fit. 😃 

[^1]: either / or, e.g. if `statistical_geogaphy:N09000004`, and `gss:S12000045` exist, render them both. But if only one, or the other, exist - handle as you would expect.

[^2]: the beta redirects _for now_, but we've mixed experience of this in the past.

lib/views/public_body/_more_info.html.erb
gbp: This looks good to me.

PR #1996, opened 2025-04-29
Accessibility improvements

lucascumsille:
## Relevant issue(s)

Fixes: https://github.com/mysociety/sysadmin/issues/2036

## What does this do?

Accessibility improvements. Most of the issues found were elements shared across all pages on WDTK.

## Why was this needed?

Improve UX and comply with The Illegal Content Codes of Practice

## Implementation notes

I used the following tools:

- ARC Toolkit
- axe Devtools

Bear in mind they still display errors, like not being able to check contrast ratios. I checked and we pass all the AA, but they are failing the AAA. Let me know if you'd like me to tackle those. I would basically have to darkened the blue in the navigation bar.

Beside that all the fixes in this commit should fix any navigation issues. The alaveteli issues were fixed here:

https://github.com/mysociety/alaveteli/pull/8675

## Screenshots

Screenshot 2025-04-30 at 06 59 59

### DIFF
--- a/app/assets/stylesheets/responsive/custom.scss
+++ b/app/assets/stylesheets/responsive/custom.scss
@@ -15,7 +15,7 @@ $color_white: #ffffff;
 $color_pro_blue: #5A7687;
 
 $mysociety-banner: $color_black;
-$link-color: darken($color_blue, 10%);
+$link-color: #0c538e; // Pass AAA Test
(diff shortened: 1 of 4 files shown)


zarino (approved):
This looks good to me! :+1:

@gbp are you happy to take it from here?

Issue #2015, opened 2025-06-23
Display warning about email/phone redactions to requester when viewing their own threads


zarino:
When a response from a public body contains an email address or phone number, Alaveteli (core) attempts to redact those details, to prevent spam.

The Alaveteli front-end does then [replace those redacted details with a link to `/help/officers#mobiles`](https://github.com/mysociety/alaveteli/blob/develop/app/models/incoming_message.rb#L519), but it can’t do that in attachments, and anyway, it’s a very subtle link, which user support messages suggest many people (fairly) miss or don’t understand – they often end up emailing the authority asking why the details have been redacted, when really it’s WDTK that’s done the redaction. Ugh.

Anyway, while we work out a better way to handle these redactions, I’m proposing a temporary improvement that at least makes it clearer what’s going on.

When the requester is looking at one of their own request threads and the final item in the thread is an incoming message, we’ll display this message, between the final response and the Reply button:

![Image](https://github.com/user-attachments/assets/7c918746-efda-4682-997c-874334f35321)

My reasoning is that maybe this way we’ll catch the user before they send their “why have you redacted the email/number in your response” follow-up to the authority.

The "Find out why" link would go to https://www.whatdotheyknow.com/help/officers#mobiles

I’d suggest we also hide the "We work to defend the right to FOI for everyone" alert when this new redaction alert is displayed, to prevent the two alerts battling for prominence.

It’s adding more visual cruft to the page, but only for requesters looking at their own requests, and only when the last activity on the request was that a response was received, so hopefully the balance of mess to utility is fair.

PR #2019, opened 2025-07-03
Added note about masked content to requester

lucascumsille:
## Relevant issue(s)
Fixes: https://github.com/mysociety/whatdotheyknow-theme/issues/2015

## What does this do?
The message advices the requester that emails and phone numbers are automatically redacted from responses. This should help reducing the amount of emails that the authorities receive from requesters asking why those details have been redacted.

## Implementation notes
@gbp 

- Initially I added this conditional`<% if @is_owning_user %>`, but I wonder if we also should include another argument that only display the message if there is a response.

- Ideally we don't want to display the "We work to defend the right to FOI for everyone" message at the same time as the warning message. I have included a conditional `<% if show_donation_button? && !flash[:request_sent] && !@is_owning_user %>`, but this conditional might change depending on the point above.

## Screenshots
Screenshot 2025-07-03 at 13 46 36

### DIFF
--- a/app/assets/stylesheets/responsive/_mixins.scss
+++ b/app/assets/stylesheets/responsive/_mixins.scss
@@ -92,6 +92,23 @@
   }
 }
 
+@mixin button-green {
+  @include button-base;
+  background: $color_green_aa; // Adding a colour instead of transparent, that way we can use this button even in dark backgrounds
+  color: $color_white;
+  font-weight: normal;
+  border-color: $color_green_aa;
+  &:hover,
+  &:active,
+  &:visited:hover,
+  &:visited:active {
+    color: $color_green_aa;
+    // We are darkening the background and border for cases there is a default button and a tertiary button together. In this case when we are hovering over the tertiary one it won't look just like the default button next to it.
+    background: $color_white;
+    border-color: $color_green_aa;
+  }
+}
+
 @mixin button-disabled {
   opacity: 0.333;
   &:hover,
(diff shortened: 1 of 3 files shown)


gbp:
Nice one @lucascumsille. I'm not sure about the conditionals but maybe @zarino might have some thoughts.

I think with the current changes the donation block isn't rendered for the requester at all so we need to consider if we want to display these messages at the same time as each other, I personally think we should show the donation block. Also @lucascumsille suggests we could, "include another argument that only display the [new redaction warning] message if there is a response" this seems sensible to me too.

garethrees:
> (I wonder whether people might skip over the red banner between messages? eg: Do we send email alerts that automatically jump down to the last message?)

Ah yes, we do and I think that would end up getting missed.

> Love the idea of putting it inside the last message!

Yeah looking at them side by side that's what I'd go for I think.

lucascumsille:
- I ended up creating a version that mixes @garethrees and @mdeuk solutions. The message is still inside the incoming message, but now I have moved it to the bottom as @mdeuk suggested. I think it makes sense considering the top side of the message already has a prominent header, so probably the warning message would have been missed.

- One thing I noticed and I changed in this PR is the donation CTA, at first glance I thought it was an error message because of the red colour. I think using green looks more positive.(@garethrees let me know if you like this change and if you prefer me to move it to another PR) 

Screenshot 2025-07-30 at 14 17 25

Screenshot 2025-07-30 at 10 41 06

garethrees:
All looks great, thanks @lucascumsille!

PR #2020, opened 2025-07-09
Add UserCheck.com integration with UserSpamScorer

gbp:
## Relevant issue(s)

Depends on https://github.com/mysociety/alaveteli/pull/8753

## What does this do?

Add UserCheck.com integration with UserSpamScorer

## Why was this needed?

Check for disposable, risky and invalid MX records for email domains.

### DIFF
--- a/lib/alavetelitheme.rb
+++ b/lib/alavetelitheme.rb
@@ -56,7 +56,8 @@ end
   'data_breach.rb',
   'excel_analyzer.rb',
   'authority_only_response_gatekeeper.rb',
-  'raw_email_usage.rb'
+  'raw_email_usage.rb',
+  'user_check_integration.rb'
 ].each { |patch| require theme_root.join('lib', patch) }
 
 $alaveteli_route_extensions << 'wdtk-routes.rb'
--- a/lib/user_check_integration.rb
+++ b/lib/user_check_integration.rb
@@ -1,0 +1,116 @@
+##
+# UserCheck.com Integration for Alaveteli UserSpamScorer
+# This module provides integration with the UserCheck.com API for email domain
+# validation
+#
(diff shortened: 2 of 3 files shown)


lib/usercheck_integration.rb
```diff
@@ -0,0 +1,130 @@
...
+      response = http.request(request)
+
+      case response.code
+      when '200'
+        data = JSON.parse(response.body)
+        {
+          success: true,
+          valid: data['valid'] != false,
+          disposable: data['disposable'] == true,
+          risky: data['risky'] == true || data['status'] == 'risky',
```
dracos, line 68: I don't see `risky` as a key in the API? And status is an HTTP status code.

lib/usercheck_integration.rb
```diff
@@ -0,0 +1,130 @@
...
+      request['Content-Type'] = 'application/json'
+
+      response = http.request(request)
+
+      case response.code
+      when '200'
+        data = JSON.parse(response.body)
+        {
+          success: true,
+          valid: data['valid'] != false,
```
dracos, line 66: valid isn't in the API, looking at https://www.usercheck.com/docs/api/domain-endpoint

dracos (changes requested):
This doesn't seem to match the API we're using, sorry if I've missed something.
On FMS we're only using disposable or blocklisted being set, low score for no MX makes sense, not sure what the mid-risky one is related to. There's "spam"  if that's the middle one?

lib/usercheck_integration.rb
gbp: Ugh sorry I don't know what happened with the responses I thought we were getting. I have pushed a fixup with hopefully match real examples.

PR #2028, opened 2025-07-31
[WIP] Welcome page to House Rules

lucascumsille:
## Relevant issue(s)
Fixes part of: https://github.com/mysociety/alaveteli/issues/8672

- [ ] @gbp I tried getting rid of the navigation bar, but I think I'll need your help with that.
- [x] @MyfanwyNixon kindly will provide a better copy for the intro, so it's more welcoming and informative.

## What does this do?

- Includes a component called `read-container` that encourages users to read its content before continuing. The button will be disabled till they finish scrolling.
- Add welcome page, where the rules are wrapped inside a `read-container` so the user can't continue without reading the whole list.

## Screenshots

https://github.com/user-attachments/assets/3d80ed27-61de-46b3-a605-85709f111bb2

### DIFF
--- a/lib/views/user/rate_limited.html.erb
+++ b/lib/views/user/rate_limited.html.erb
@@ -17,7 +17,7 @@
 	        "please <a href='{{help_contact_path}}'>get in touch</a>.",
 	     :help_contact_path => help_contact_path) %></p>
 
-<% if feature_enabled?(:alaveteli_pro) %>
+<% if feature_enabled?(:alaveteli_pro) && !user.has_tag?('disposable_email') %>
   <div class="pro-upsell">
     <div class="pro-upsell__headers">
       <div class="pro-upsell__header">


lucascumsille:
@garethrees I just added the new intro provided by Myf.

Screenshot 2025-07-31 at 14 51 22


Myf's original included also the following paragraph:
> Note that if you break a rule, we may suspend your account.

But because we have the warning at the bottom I thought it wasn't needed.

garethrees:
Looks great!

garethrees:
> I'm not a big fan of having two pieces of important text side by side to avoid competition between content I thought adding a carousel, so users have some time to digest information.

Nice, yeah I like this – it's a better way of providing a moment to pause and gives one bit of info at a time.

> I was thinking using a Pop up

I feel a bit more mixed about this. I like that it reuses code so reduces a bit of complexity there, but I do feel that as soon as I see a modal my instinct is to mash `esc` to make it go away because it's probably trying to upsell me.

From the video it's not clear how this is triggered – presumably from the "or start a batch request" link? And then I guess we just need to make the "Next →" button on the second page of the carousel go through to the batch interface?

Maybe we could use the modal pattern, but make it fill out the page a bit more, and not have quite as dark a background overlay – make it _feel_ more like an interstitial page, even though it's implemented as a modal?

> For example when using a Pop up where we don't want users to easily dismiss it and close it by just clicking outside the pop up we can use `data-click-outside-close="false"` so the only way to close the pop up is by pressing the close button.

I don't think we should prevent people closing it however they prefer – I just think we should prevent access to batch unless they click through the carousel.

> I have summarise the rules and include a link to the full version in case users want to go there.

I actually think that given this design we could feasibly render the full house rules on this second page, as we do on the "welcome" interstitial – especially if we made the "modal" look a bit more like an interstitial? That would save maintaining two versions, which would be a big win.

PR #2029, opened 2025-08-04
Improve styling of FOI Myths on Learn page

zarino:
Headings elsewhere in the Learn materials state facts. Which made this myths page confusing, as you instinctively want to trust what’s marked up as a heading, but here, the headings were all lies!

Screenshot 2025-08-04 at 16-26-50 Common misconceptions about FOI - WhatDoTheyKnow

This commit introduces new styling which wraps the myths in a red box (to imply “wrongness”) and instead uses the “correction” immediately after each myth as the heading, to avoid misinterpretation when skim-reading the page.

Screenshot 2025-08-04 at 16-26-56 Common misconceptions about FOI - WhatDoTheyKnow

### DIFF
--- a/app/assets/stylesheets/responsive/learn.scss
+++ b/app/assets/stylesheets/responsive/learn.scss
@@ -118,6 +118,27 @@ $card-transition: 0.35s cubic-bezier(0.30, 1, 0.7, 1);
     }
   }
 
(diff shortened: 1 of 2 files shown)


garethrees:
Nice improvement yeah. _Love_ the little tilt on the label.

PR #2034, opened 2025-08-19
Add email domain report mailer

gbp:
## What does this do?

Compare which email domains are being used for new user registrations. Showing a percentage change over the preceding month. Required at least two registrations to have happened for a domain to be included.

Can be delivered with:
  ./bin/rails runner 'UserMailer.email_domain_report.deliver_now'

## Screenshots

image

### DIFF
--- a/lib/alavetelitheme.rb
+++ b/lib/alavetelitheme.rb
@@ -58,7 +58,8 @@ end
   'authority_only_response_gatekeeper.rb',
   'raw_email_usage.rb',
   'immigration_detection.rb',
-  'user_check_integration.rb'
+  'user_check_integration.rb',
+  'email_domain_report.rb'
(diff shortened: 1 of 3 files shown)


garethrees (commented):
Had a quick scan through and looks good other than a potential issue with the email formatting.

I think we should ship this in core unless you think there's a really compelling reason not to? If we do so, I wouldn't add it under `UserMailer` as that's a bit more about transactional mail to users rather than stuff to _admins_.

Could make it on/off configurable at theme level – something like:

```rb
class Admin::SignupMailer
  cattr_accessor :enabled, default: true

  def weekly_domain_report
    return false unless enabled?

    # …
  end
end

# Disable in theme if you really want to:
# Admin::SignupReportMailer.enabled = false
```

Issue #2040, opened 2025-09-05
Update privacy policy with Stripe cookie details


garethrees:
WDTK loads StripeJS when pro users browse their “Subscriptions” page. This, in turn, loads an hCaptcha - which sets a cookie.

Need to add this to the cookies list.

garethrees:
## Current state on WhatDoTheyKnow

We have two places of relevance:

* [`/help/privacy`](https://www.whatdotheyknow.com/help/privacy): We don't mention Stripe here.
* [`/pro/pages/legal`](https://www.whatdotheyknow.com/pro/pages/legal): We do mention Stripe here.

> We use Stripe for processing payments. You can read about Stripe’s usage terms and privacy policy here: https://stripe.com/gb/privacy.
>
> – https://www.whatdotheyknow.com/pro/pages/legal

## What do we do elsewhere?

### MapIt

> What personal data we collect
> …
> If you make a purchase, your name, contact details, payment and security details are held by the payment processor Stripe, whose usage terms and privacy policy can be seen here: [stripe.com/gb/privacy](https://stripe.com/gb/privacy).
> 
> Your name, email address, the billing address of your card and the last few digits of your card or bank account number are made available to us along with your purchase history.
>
> – https://mapit.mysociety.org/privacy/

### TheyWorkForYou

> If you register as a paying customer for the API, your name, contact details, payment and security details are held by the payment processor Stripe, whose usage terms and privacy policy can be seen here: [stripe.com/gb/privacy](https://stripe.com/gb/privacy).
> 
> Your name, email address, the billing address of your card and the last few digits of your card or bank account number are made available to us along with your purchase history. We use these to issue invoices and send you important information about changes or updates to the service.
>
> – https://www.theyworkforyou.com/privacy/

PR #2044, opened 2025-09-16
[#2040] Update Stripe processing details

garethrees:
Update in line with other mySociety services.

Fixes https://github.com/mysociety/whatdotheyknow-theme/issues/2040.

Appears on `/pro/pages/legal`. Not added to `/help/privacy` as not relevant unless you're a Pro subscriber.

Screenshot 2025-09-16 at 14 38 36


SarahmySociety:
Great stuff, good to make consistent with other services. We may want to revisit wording later to check there's no duplication with the main privacy policy, but that won't block this merge.

PR #2052, opened 2025-09-19
Refactor UK-specific webmail providers

garethrees:
Depends on https://github.com/mysociety/alaveteli/pull/8889

We now have a global listing of webmail providers [1], which is used in various places where we use email address domains to do something.

We can now extract the UK-specific domains from
`PublicBody.excluded_calculated_home_page_domains` to this webmail providers list as they are all domains of that nature. The calculated home page domains list will inherit the customised webmail provider list.

[1] https://github.com/mysociety/alaveteli/commit/c0764b7

### DIFF
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -32,6 +32,24 @@ Rails.configuration.to_prepare do
                      "to government services.")
   end
 
+  # Add UK-specific domains and providers
+  Domains.webmail_providers += %w[
+    aol.co.uk
+    blueyonder.co.uk
+    btconnect.com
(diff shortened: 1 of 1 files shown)


garethrees:
Ah annoying. Looks like the `PublicBody.excluded_calculated_home_page_domains` default (and similarly `PublicBody.excluded_foi_officer_access_domains`) is already initialised prior to setting our custom `Domains.webmail_providers`.

@gbp any idea of what we could do here? Ideally would like to avoid having to explicitly include the updated webmail providers.

PR #2056, opened 2025-09-22
Reconcile with core

garethrees:
See https://github.com/mysociety/alaveteli/pull/8891. Fix issue with form_item_note being jammed up against the signup button.

**BEFORE**

Screenshot 2025-09-22 at 16 08 15

**AFTER**

Screenshot 2025-09-22 at 16 08 01


garethrees:
Build failing because core branch with same name now destroyed after merge. Just going to merge as it's only a view reconcilliation.

PR #2058, opened 2025-09-26
Add translation disclaimer

garethrees:
We don't have the resources to keep translations in sync, so these often go out of date. In some cases it's quite important that the reader knows that the English version is the definitive one.

Fixes https://github.com/mysociety/whatdotheyknow-private/issues/417.

- [ ] Would be good to translate the actual disclaimer to Welsh.

Have added this to all non-default locale help pages, other than the contact form which is a bit of a different case as it's not really "help" content as such.

Screenshot 2025-09-26 at 12 23 09

### DIFF
--- a/lib/views/help/privacy.cy.html.erb
+++ b/lib/views/help/privacy.cy.html.erb
@@ -1,22 +1,11 @@
 <% @title = "Eich preifatrwydd" %>
 <%= render :partial => 'sidebar' %>
+
 <div id="left_column_flip" class="left_column_flip">
+  <%= render partial: 'translation_disclaimer' %>
+
   <h1 id="privacy"><%= @title %> <a href="#privacy">#</a> </h1>
   <dl>
-    <div class="changes">
-      <span style="font-size:16px">
-      <p>
-        <strong>Pennau i fyny!</strong> Rydym yn gwybod nad yw'r dudalen hon mor gyfoes ag y dymunwn. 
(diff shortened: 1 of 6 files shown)


lib/views/help/privacy.cy.html.erb
```diff
@@ -1,22 +1,11 @@
...
   <dl>
-    <div class="changes">
-      <span style="font-size:16px">
-      <p>
-        <strong>Pennau i fyny!</strong> Rydym yn gwybod nad yw'r dudalen hon mor gyfoes ag y dymunwn. 
-        Rydyn ni'n ceisio datrys hyn; ond os na allwch ddod o hyd i'r hyn sydd ei angen arnoch yma, 
-        rhowch gynnig ar ein prif <a href="/help/privacy">Hysbysiad Preifatrwydd</a>, sydd yn Saesneg.
-      </p>
-      <p>
-        Os hoffech chi, gallwch weld cyfieithiad awtomataidd, drwy ddefnyddio
```
mdeuk, line 14: It would be good to maintain this link, in some way - as it allows for an easy way to obtain the content in a language other than `en_GB`.

garethrees: Ah yeah okay ISWYM

garethrees: Added in 55fc54c

PR #2075, opened 2026-02-05
Scrappy A/B test of donation landing pages

garethrees:
Temporary override to the configuration variable to A/B test conversion of the service-specific donation page and the general organisation one.

This won't necessarily be a true 50:50 split (since `sample` is random and the various caching layers may have an impact, etc) but its the landing:conversion rate we're mostly interesting.

### DIFF
--- a/lib/model_patches.rb
+++ b/lib/model_patches.rb
@@ -19,6 +19,13 @@ Rails.configuration.to_prepare do
       AlaveteliSpamTermChecker::DEFAULT_SPAM_TERMS + custom_terms
   end
 
+  AlaveteliConfiguration.instance_eval do
+    def donation_url
+      ['https://www.mysociety.org/donate/support-whatdotheyknow-and-mysociety/',
+       'https://www.mysociety.org/donate/'].sample
+    end
+  end
+
   ContactValidator.class_eval do
     attr_accessor :understand
 
--- a/spec/donation_helper_spec.rb
+++ b/spec/donation_helper_spec.rb
@@ -7,6 +7,10 @@ RSpec.describe DonationHelper, type: :helper do
     subject { show_donation_button? }
 
     context 'without donation URL' do
+      before do
+        allow(AlaveteliConfiguration).to receive(:donation_url).and_return('')
+      end
+
       it { is_expected.to eq false }
     end
 


lucascumsille:
Hi @alicewilliams36 and @garethrees 

I made the following page:

https://github.com/user-attachments/assets/5b24055a-bdae-4984-8380-9a6cc86cec90

Let me know if you'd like me to make any changes. Btw please ignore the Mechelen picture, I couldn't find the one that Alice used.

alicewilliams36:
This is fantastic thankyou Lucas! Just a couple of edits:
1. Could you swap the wording 'Make Information Accessible To All' with 'Join the mySociety Supporter Network' so that the former is the larger title, and the latter is in the box above?
2. Here are 2 options of images instead of Mechelen https://drive.google.com/drive/folders/1xXgDAKLMyxWdUUSN1TkSoNOUqfPA5xHo?usp=drive_link 
3. Is there any chance the text next to that image (starting 'when information is hidden') could be any larger?
4. Do the donate buttons jump you down to the form? Could the button below the 3 boxes be a full width button? And the same for the 20 year impact button?
5. Sorry now I've seen the scroll I think we need to move Maya's quote to below the form so you don't have to scroll too far to get to that.
Thanks so much!

lucascumsille:
Thank you @alicewilliams36 =)
No probs I'll make those changes now, also if after that you are happy for me to deploy I can do.
I still need to add some javascript for the buttons though.

lucascumsille:
Here is a new version with the latest feedback:

https://github.com/user-attachments/assets/0604fb00-43ca-43f0-914c-f1487c6ca3ca

Let me know if you'd like me to change anything else.

lucascumsille:
Thanks for all the feedback from yesterday, the new page is [live here](https://www.mysociety.org/donate/support-whatdotheyknow-and-mysociety/)

Let me know if there is anything else I need to do

garethrees:
Will have a proper read through as soon as I can but looks way nicer! One thing I did just spot:

> Provide support for the other countries using our code to run their own FOI site – from Jordan to Ukraine

There's no Alaveteli in Jordan https://alaveteli.org/deployments/

alicewilliams36:
Oops! Sorry @lucascumsille - could you edit that to say 'from Ukraine to Colombia' (Trying to highlight the range whilst also naming countries people will care about!)

lucascumsille:
@alicewilliams36 I just deployed that change =)

garethrees:
This is looking great. Got two minor suggestions.

1. The "Donate £5 monthly" button in the header pre-fills the donation form with the correct options. "Other amount" doesn't. Would be nice if it did if it's an easy fix, but wouldn't worry about it if it would take hours.
2. Might be worth calling out a key phrase in the first paragraph to make it a little more scannable? One of these perhaps? Don't feel strongly about this at all though.

Screenshot 2026-04-02 at 09 45 39

Screenshot 2026-04-02 at 09 46 43

garethrees:
@gbp can you have a quick look at the code side of this (i.e. this PR)? TL;DR is that we want to A/B test the performance of the general and service-specific donation pages. I've hacked this in by just overriding `AlaveteliConfiguration#donation_url` with hard coded values and sampling from them. Think this should be fine, but LMK if you think there would be issues. Only problem I can think of is that it wouldn't be obvious that the configured value in `config/general.yml` would get ignored if we were to change it.

alicewilliams36:
@lucascumsille Just checking if you saw those little edit suggestions from Gareth? On making the text bold I think that's a good idea and think option 2 above is the best. (**Your donation...**)

On the donation form, relooking at that could we please change the amounts to £5, £10 and £40? And on prepopulating the form', if it's easy to prep-populate 'other amount' to £10 monthly then great but not to worry if that's a complex change as Gareth says!

garethrees:
> On the donation form, relooking at that could we please change the amounts to £5, £10 and £40?

Any reason to deviate from the main donate page? I'd have thought that might skew the results a bit (though tbf probably close enough that its not a huge concern)

lucascumsille:
@alicewilliams36 and @garethrees thank you both I just made the changes requested:

- Bold -> `Your donation helps ensure that information remains accessible to everyone — not just the powerful. `
- Now "Other amount" prefills £10 monthly)

Regarding:

> On the donation form, relooking at that could we please change the amounts to £5, £10 and £40?

Let me know if you'd rather stick to what we currently have or display these options instead. I think it should be easy to change, but not super sure, because I haven't done it before. It might be just changing the array for donations.

alicewilliams36:
Thanks so much Lucas! Yes please, I've done a bit of research into donation amounts so think offering £5, £10 and £40 will be more fruitful than the current options!

lucascumsille:
@alicewilliams36 After our talk from yesterday I have updated the donations to 5,10,40 for WDTK and Support mySociety, I also added [Climate](https://www.mysociety.org/donate/support-mysociety-climate-programme/) because I think it would be a bit odd to be the only one with the past values.

alicewilliams36:
Makes sense thankyou! I'm not understanding the github notes above re the unsuccessful check - is that where @garethrees is actioning the testing function?

garethrees:
> I'm not understanding the github notes above re the unsuccessful check - is that where @garethrees is actioning the testing function?

Its some automated testing – the breakage is unrelated to this change. Will resolve.

alicewilliams36:
@lucascumsille sorry I had some feedback at the team meeting that the filing cabinet image at the top isn't quite working... would you be able to switch it out for a people image? Perhaps [this one](https://drive.google.com/file/d/19V4kqoCD8qh4uL4UVaQ2GM9qSACNJKT7/view?usp=sharing)

garethrees:
@alicewilliams36 fyi the A/B test is now live – anywhere a donation link is shown one of the variants is picked at random.

alicewilliams36:
Fantastic thankyou Gareth! Looking forward to seeing how this does :)

lucascumsille:
@alicewilliams36 I just updated the image to the proposed one, because the image ratio is a bit tricky I had to increase the vertical padding of the hero section.

PR #2121, opened 2026-07-01
Add account management page

### LINKED ISSUE #795: [alaveteli/6169] Create a /help page regarding account security / two factor authentication
We should create a page in /help that provides users with information about how to keep their account safe, and how to use two factor / multi factor authentication.

I'm splitting this off from the master ticket so we can look at any WDTK specific scenario.

@MattK1234  wrote in mysociety/alaveteli#6169:

> 
> 
> image
> 
> We don't ask for the two factor passcode particularly often. Occasionally the WDTK team receive some correspondence from users who might have forgotten about this.
> 
> As 2FA can take many forms some users are expecting a text message or email address or some other form of communication.
> 
> We could provide some additional information here to help users understand what we are asking for, maybe by a little question mark of help icon displaying the some text or similar.
> 
> We could also highlight our inability to do anything if the OTP is forgotten.
> 
> Is it really a one-time passcode though? Surely it would be used every time a user changes their password? We might want to consider changing that terminology.

FOIMonkey:
## Relevant issue(s)
#795 
## What does this do?
Adds basic detail of 2fA
## Why was this needed?
We had nothing about it
## Implementation notes
n/a
## Screenshots
Screenshot 2026-07-01 at 16 43 06
## Notes to reviewer
Initial attempt - feel free to improve :)

### DIFF
--- a/lib/config/wdtk-routes.rb
+++ b/lib/config/wdtk-routes.rb
@@ -51,6 +51,9 @@ Rails.application.routes.draw do
   get '/help/beginners' => 'help#beginners',
       as: :help_beginners
 
+  get '/help/account_management' => 'help#account_management',
+      as: :help_account_management
+
   get '/help/ico_officers' => 'help#ico_officers',
       as: :help_ico_officers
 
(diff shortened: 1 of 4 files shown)


garethrees:
perfect thanks!

PR #2124, opened 2026-07-06
Reconcile with core

gbp:
`CensorRule#info_request`/`CensorRule#public_body` -> `CensorRule#censorable`

Required for core#9302


garethrees:
argh sorry should have caught that – thanks

PR #2125, opened 2026-07-08
Add Public Body name to analyzer report

garethrees:
It's useful to know which authority the request was sent to as this gives a lot of context to the data that the file may contain.


garethrees:
@gbp trival change but just want to check that there wasn't an explicit reason for not including this?

mdeuk:
Just an observation - but, would including the request email address, and `subject` from the incoming email, not be useful here?

garethrees:
I'm actually working on some scripts (just running them on server at the moment) to automate a lot of the handling. One thing it does is outputs envelope details (to and subject, based on the request subject) and then generates the letter, which includes request email and such (see inbox sent folder for some recent examples).

We didn't include subject/request title in the analyser report due to potential of PII.

gbp (approved):
See no reason not to add this

PR #2134, opened 2026-07-24
[private#428] Refactor house rules

garethrees:
* Add general top-level rule about using the service in line with its purpose
* Group rules by logical areas (general conduct, requesting, accounts)
* Compress similar rules into a single line
* Phrasing tweaks
* Intentionally use unordered lists. We need to be able to update these from time to time and don't want to be constrained by previous rule numbering where we've referred to a specific rule by number.

Fixes https://github.com/mysociety/whatdotheyknow-private/issues/428.

Before and after:

Screenshot 2026-07-24 at 16 00 01

### DIFF
--- a/lib/views/help/house_rules/_list_of_house_rules.html.erb
+++ b/lib/views/help/house_rules/_list_of_house_rules.html.erb
@@ -1,25 +1,117 @@
-  <ol>
-    <li>Only use WhatDoTheyKnow to request specific information. Don't use it to send general messages to public authorities or for personal messages.</li>
-    <li>Don't include things that could be seen as harmful or untrue in your requests and comments. Don't make allegations.</li>
-    <li>Don't post anything that is against the law, threatening, abusive, or offensive. Any such content will be <%= link_to 'removed', help_removing_information_path %>.</li>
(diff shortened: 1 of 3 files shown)


lib/views/help/house_rules/_list_of_house_rules.html.erb
```diff
@@ -1,25 +1,117 @@
...
+
+  <li>
+    <strong>Don’t make requests that have no real purpose or are meant to cause
+    problems</strong> for a public body. The ICO has
+    <a href="https://ico.org.uk/for-organisations/foi/freedom-of-information-and-environmental-information-regulations/section-14-dealing-with-vexatious-requests/">
+    guidance on this type of vexatious request</a>.
+  </li>
+
+  <li>
+    <strong>Don’t ask for email addresses in your requests</strong>. Email
```
garethrees, line 81: I was originally going to just move this to `/help/requesting#responsible`, but decided against it at the last minute. Its the right logical place, but I think that help page is much too long at the moment. This is quite an important rule, so keeping it here until we look at the requesting page.

